PR #1, opened 2017-01-18
Add #![no_std]

tonychain:
Use `::core` in lieu of `::std`, allowing this crate to be usable in `#![no_std]` environments.

Gates features that presently depend on ::std (presently just rand) behind a "std" cargo feature, which is enabled by default.

Switches to the `sha2` crate in lieu of `rust-crypto`: the `sha2` crate is part of an effort to break up `rust-crypto` from a monolithic library into a set of interoperable crates, but is otherwise mostly just `s/crypto:://`. The `sha2` crate already supports `#![no_std]` out of the box (unlike `rust-crypto`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ exclude = [ ".gitignore", "TESTVECTORS" ]
 arrayref = "0.3.2"
 rust-crypto = "^0.2"
 rand = "^0.3"
-curve25519-dalek = "^0.1"
+curve25519-dalek = "^0.3"
 
 [dev-dependencies]
 rustc-serialize = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,20 @@ exclude = [ ".gitignore", "TESTVECTORS" ]
 
 
 [dependencies]
-arrayref = "0.3.2"
-rust-crypto = "^0.2"
-rand = "^0.3"
-curve25519-dalek = "^0.3"
+arrayref = "0.3.3"
+sha2 = "^0.4"
+
+[dependencies.curve25519-dalek]
+version = "^0.3"
+default-features = false
+
+[dependencies.rand]
+optional = true
+version = "^0.3"
 
 [dev-dependencies]
 rustc-serialize = "0.3"
+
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -12,9 +12,16 @@ exclude = [ ".gitignore", "TESTVECTORS" ]
 
 [dependencies]
 arrayref = "0.3.2"
-rust-crypto = "^0.2"
-rand = "^0.3"
-curve25519-dalek = "^0.1"
+sha2 = "^0.1"
+curve25519-dalek = { git = "https://github.com/chain/curve25519-dalek", branch = "no-std" }
```
tonychain: Will need to merge this first:

https://github.com/isislovecruft/curve25519-dalek/pull/17

tarcieri:
I just confirmed the current PR actually builds within a `#![no_std]` program, so it should be safe to merge after https://github.com/isislovecruft/curve25519-dalek/pull/17

src/ed25519.rs
```diff
@@ -129,7 +129,7 @@ impl SecretKey {
     /// A `SecretKey`.
     #[inline]
     #[allow(dead_code)]
-    fn from_bytes(bytes: &[u8]) -> SecretKey {
+    pub fn from_bytes(bytes: &[u8]) -> SecretKey {
```
tarcieri: I imagine you'll have to do something like this eventually, otherwise there's presently no way to load an existing secret key, only generate a random one.

I believe in @briansmith's *ring* library he recomputes the Ed25519 public key from the private key to ensure it matches (and therefore the keypair is valid)

tarcieri: Oh, and now I noticed `SecretKey` is `pub struct SecretKey(pub [u8; 64])`. I guess `SecretKey(bytes)` works but I think I'd prefer some sort of constructor function.

tarcieri:
This is now updated to use the released curve25519-dalek v0.3.0 crate, and should be ready to merge.

isislovecruft:
Thanks a bunch, @tarcieri! I merged last week and released version 0.2.0.

PR #3, opened 2017-03-13
Add conditional compilation to allow use on stable

tqoitc:
Same as PR submitted on curve25519-dalek except also adds a `cfg_attr` on the "rand" feature. That should be changed to require the `rand` crate on crates.io.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,9 +59,9 @@
 //! ```
 
 #![no_std]
-#![feature(rand)]
+#![cfg_attr(std, feature(rand))]
 #![allow(unused_features)]
-#![feature(test)]
+#![cfg_attr(test, feature(test))]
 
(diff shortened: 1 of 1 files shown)


burdges:
Is there ever a reason to use the rand feature instead of the `rand` crate?  See https://github.com/rust-lang/rust/issues/27703 https://github.com/rust-lang/rust/issues/36999 https://internals.rust-lang.org/t/rand-crate-brainstorming-minimal-api-in-std/3731

isislovecruft:
Hi, @tqoitc! Thanks for the patch! I'm sorry for only just noticing this PR.

`src/lib.rs` was changed in 4eaa1321 and 85b0cd93 and I believe the new behaviour should work fine on stable. Please let me know if that's not the case.

PR #15, opened 2017-12-06
Use a custom error type instead of &'static str.

withoutboats:
Advantages of a custom error type:

- It can be more easily integrated into other error types by clients;
  they can implement From<FromBytesError> for their error types, or
  they can use a library like failure.
- It is a zero-sized type, which can enable some representational
  optimizations.
- It can be easier and more stable to test for.

However, this is a question here. By using a single `FromBytesError`, each
error loses the contextual information about what you were parsing from bytes.
We could instead use a custom error type for each function (`SignatureError`,
`PublicKeyError`, etc) if this contextual information seems valuable.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,10 @@ optional = true
 version = "^0.6"
 optional = true
 
+[dependencies.failure]
+version = "^0.1.1"
+default-features = false
+
 [dev-dependencies]
 hex = "0.2"
 sha2 = "^0.6"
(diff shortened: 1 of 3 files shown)


pwoolcoc:
ha, I was just coming to this repo to see if this had been proposed or discussed in the issues yet, and here I see withoutboats has beat me by 2 days :)

isislovecruft:
Hi @withoutboats!

Thanks for the PR! This is great! As is likely obvious, I've not dealt with error handling at all yet, so the help is very appreciated. I'll gladly merge it.

A couple thoughts:

1) As you mentioned, I think there should be somewhat more robust/descriptive error types. So, for example, because a `Signature` is `(R, s)` where `R` is a compressed curve point and `s` is a scalar, there might be a `SignatureDecompressionError`** in the public API which has two underlying error types causing it, `ExtendedPointDecompressionError` and `ScalarFormatError`, each with different messages.
2) We might want to extend these error/fail types into curve25519-dalek, cc @hdevalence.

PR #16, opened 2017-12-12
Do not require feature=std for PublicKey::from_secret

oleganza:
In my use case I generate a secret key at one point in time, and then have to generate a corresponding public key later. The 0.5.0 version of ed25519-dalek has a leftover `feature="std"` requirement that's no longer needed.


isislovecruft:
Hi @oleganza! Thanks, I've merged this for 0.6.0.

PR #26, opened 2018-07-01
Upgrade to rand 0.5 (fixes #23)

### LINKED ISSUE #23: Request: Move to rand 0.5.
The release of `rand 0.5` brought some major changes and it would be great to have an ed2519-dalek release which is compatible with these changes.

The current master branch still depends on `v.04` but the development branch already got an update to to a pre-release of `v0.5`.  Is there any additional work required to move this to the actual release version and merge that changes into master?

tarcieri:
This switches `Keypair::generate` from a trait object to an explicit type parameter, which unfortunately I think is unavoidable given the new `rand` API.

### Bounded trait object

Doesn't work 😢 

```
error[E0225]: only auto traits can be used as additional traits in a trait object
   --> src/ed25519.rs:313:47
    |
313 |     pub fn generate(csprng: &mut (CryptoRng + RngCore)) -> SecretKey {
    |                                               ^^^^^^^ non-auto additional trait
```

### Phantom data parameter on `Keypair`

This is probably still worth considering, although it's a bit awkward. The RNG to use in conjunction with `::generate` could be provided as a type parameter on `struct Keypair` itself, which would allow supplying a default (i.e `struct Keypair<R: CryptoRng + RngCore = OsRng>`), but seems a bit awkward.

I'm still confused why `CryptoRng` is not defined as `CryptoRng: RngCore` in `rand`. At one point I think I figured out why but I have since forgotten...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ default-features = false
 
 [dependencies.rand]
 optional = true
-version = "0.4"
+version = "0.5"
 
 [dependencies.digest]
 version = "^0.7"
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -13,7 +13,7 @@
 use core::fmt::{Debug};
 
 #[cfg(feature = "std")]
-use rand::Rng;
+use rand::{CryptoRng, RngCore, OsRng};
 
 #[cfg(feature = "serde")]
 use serde::{Serialize, Deserialize};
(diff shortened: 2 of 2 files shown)


0ndorio:
@tarcieri .. Out of curiosity. Is there a reason why you considered to change the signature to use an inline bounded trait but not to introduce a new trait which combines the two given ones? Seems like a breaking change anyways. 

```rust
pub trait OurRngTrait: CryptoRng + RngCore {}

pub fn generate(csprng: &mut OurRngTrait) -> SecretKey {}
```

tarcieri:
I took a shot at your suggested approach in 757ae76. Introducing an extra trait is a little annoying because it needs to be impl'd for any RNG you want to use, but 99% of the time people should be using `OsRng` anyway, and it certainly makes for a much less obtrusive change which I think is at least less annoying than introducing another type parameter.

isislovecruft:
Oh hmm, I updated it to 0.5 a couple days ago, but I didn't have this issue?  I think perhaps it's something that was fixed in either 0.5.2 or 0.5.3, and I vaguely recall having this issue with using the 0.5.0 release in another project I was working on.

In any case, this is fixed now in a4be92b! Also, sorry for not seeing your PR earlier!

PR #28, opened 2018-07-11
Switch to using criterion for benchmarks.

isislovecruft:


hdevalence:
This looks good to me. The only thing I wonder about is whether it's really worth defining and calling a `ZeroRng` instead of just using `ThreadRng`... it's extra code, the cost of the `ThreadRng` is really small, and it's further away from the keygen numbers that people actually care about (since people probably care more about generating keys with an actual RNG).

PR #30, opened 2018-07-13
Make `verify` return a `Result`.

hdevalence:
This also simplifies the verification logic.  Because the verification check
happens in variable time, we don't need to do a constant-time eq check at the
end, so we can drop the `subtle` dependency entirely.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -36,9 +36,7 @@ use curve25519_dalek::edwards::CompressedEdwardsY;
 use curve25519_dalek::edwards::EdwardsPoint;
 use curve25519_dalek::scalar::Scalar;
 
-use subtle::ConstantTimeEq;
-
-use errors::DecodingError;
+use errors::SignatureError;
 use errors::InternalError;
 
 /// The length of a curve25519 EdDSA `Signature`, in bytes.
@@ -132,9 +130,9 @@ impl Signature {
 
     /// Construct a `Signature` from a slice of bytes.
     #[inline]
-    pub fn from_bytes(bytes: &[u8]) -> Result<Signature, DecodingError> {
+    pub fn from_bytes(bytes: &[u8]) -> Result<Signature, SignatureError> {
         if bytes.len() != SIGNATURE_LENGTH {
-            return Err(DecodingError(InternalError::BytesLengthError{
+            return Err(SignatureError(InternalError::BytesLengthError{
                 name: "Signature", length: SIGNATURE_LENGTH }));
         }
         let mut lower: [u8; 32] = [0u8; 32];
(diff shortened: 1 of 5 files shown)


src/ed25519.rs
```diff
@@ -727,36 +719,27 @@ impl PublicKey {
...
-        ao = self.decompress();
-
-        if ao.is_some() {
-            a = ao.unwrap();
-        } else {
-            return false;
-        }
-        a = -(&a);
+        let A = self.0.decompress()
+            .ok_or_else(|| SignatureError(InternalError::PointDecompressionError))?;
```
isislovecruft: I think I prefer the way I did it with `match` in #31? If that's okay.

src/ed25519.rs
```diff
@@ -694,12 +692,6 @@ impl PublicKey {
         Ok(PublicKey(CompressedEdwardsY(bits)))
     }
 
-    /// Convert this public key to its underlying extended twisted Edwards coordinate.
-    #[inline]
-    fn decompress(&self) -> Option<EdwardsPoint> {
-        self.0.decompress()
-    }
```
isislovecruft: Maybe don't remove this? I'm using it in #31.

src/ed25519.rs
```diff
@@ -727,36 +719,27 @@ impl PublicKey {
...
-        let mut digest: [u8; 64] = [0u8; 64];
-
-        ao = self.decompress();
-
-        if ao.is_some() {
-            a = ao.unwrap();
-        } else {
-            return false;
-        }
-        a = -(&a);
```
isislovecruft: Why did you change to negating `k` instead?

hdevalence: No real reason, I wanted to avoid a point operation, but negating the Edwards point is probably cheaper (since it doesn't require a reduction mod L)

src/ed25519.rs
hdevalence: It's up to you, but I think this is a lot cleaner.

src/ed25519.rs
hdevalence: I think it's only used there in the same way that it was used here, so my preference would be to remove it there too.

hdevalence:
Changed title to `[WIP]`, I'll update this PR.

What do you think of the error / result changes?

isislovecruft:
The errors and results are good.

src/ed25519.rs
isislovecruft: I benched scalar and Edwards point negation and point negation is like twice as fast (they're both in the tens of nanoseconds on my machine so not that it really matters but).

hdevalence:
Rebased onto `develop`.

PR #31, opened 2018-07-13
implement ed25519ph

isislovecruft:


hdevalence:
Are there any other implementations of `ed25519ph` that this could compare against, if the RFC test vectors are wrong?

isislovecruft:
Yes, libsodium implements it, and on cursory glance [appears to do so correctly](https://github.com/jedisct1/libsodium/commit/707f4c8e085e7e2e4c12d8568ba81eaad5f6ba33#diff-ad79a080b63a6d445b935725ca40c299R13). It doesn't support adding contexts, however.

isislovecruft:
Okay the vector passes now.

src/ed25519.rs
```diff
@@ -1056,7 +1056,86 @@ impl Keypair {
...
+    /// actually a Bitcoin transaction moving all your magic internet money to
+    /// their address, it'd be a valid transaction.
+    ///
+    /// By adding a context, this trick becomes impossible, because the context
+    /// is concatenated into the hash, which is then signed.  So, going with the
+    /// previous example, if your bitcoin wallet used a context of
+    /// "BitcoinWalletAppTxnSigning" and OpenPGP used a context (this is likely
+    /// the least of their safety problems) of "GPGsCryptoIsntConstantTimeLol",
+    /// then the signatures produced by both could never match the other, even
+    /// if they signed the exact same message with the same key.
```
hdevalence: salty

src/ed25519.rs
```diff
@@ -1056,7 +1056,86 @@ impl Keypair {
...
+    ///
+    /// let sig: Signature = keypair.sign_prehashed(prehashed, Some(context));
+    /// # }
+    /// #
+    /// # #[cfg(any(not(feature = "sha2"), not(feature = "std")))]
+    /// # fn main() { }
+    /// ```
+    ///
     /// [rfc8032]: https://tools.ietf.org/html/rfc8032#section-5.1
+    /// [terrible_idea]: https://github.com/isislovecruft/scripts/blob/master/gpgkey2bc.py
```
hdevalence: this exists?????

src/ed25519.rs
isislovecruft: what are you talking about, i am the sweetest person ever

src/ed25519.rs
isislovecruft: yes lol

read the gnupg-devel mailing list for a steady stream of hilariously bad ideas

src/ed25519.rs
```diff
@@ -593,6 +593,91 @@ impl ExpandedSecretKey {
...
+    ///   which may be used to provide additional domain separation.  If not
+    ///   set, this will default to an empty string.
+    ///
+    /// # Returns
+    ///
+    /// An Ed25519ph [`Signature`] on the `prehashed_message`.
+    ///
+    /// [rfc8032]: https://tools.ietf.org/html/rfc8032#section-5.1
+    pub fn sign_prehashed<D>(&self,
+                             prehashed_message: D,
```
hdevalence: Why does `sign_prehashed` take a digest instance, instead of just taking the prehash (as a `[u8; 64]` or whatever?)

The digest instance is immediately consumed to produce the prehash, so there's no benefit internal to the function from taking a digest, but it makes the API less flexible, since the computation of the hash value is still coupled to the signing.

src/ed25519.rs
```diff
@@ -593,6 +593,91 @@ impl ExpandedSecretKey {
...
+        // still bleeding from malleability, for fuck's sake.
+        h.input(b"SigEd25519 no Ed25519 collisions");
+        h.input(&[1]); // Ed25519ph
+        h.input(&[ctx_len]);
+        h.input(ctx);
+        h.input(&self.nonce);
+        h.input(&prehash);
+        hash.copy_from_slice(h.fixed_result().as_slice());
+
+        mesg_digest = Scalar::from_bytes_mod_order_wide(&hash);
```
hdevalence: This would be cleaner as `Scalar::from_hash(h)`, and you wouldn't need the `hash` variable.

src/ed25519.rs
```diff
@@ -593,6 +593,91 @@ impl ExpandedSecretKey {
...
+        h.input(r.as_bytes());
+        h.input(public_key.as_bytes());
+        h.input(&prehash);
+        hash.copy_from_slice(h.fixed_result().as_slice());
+
+        hram_digest = Scalar::from_bytes_mod_order_wide(&hash);
+
+        s = &(&hram_digest * &self.key) + &mesg_digest;
+
+        Signature{ r: r, s: s }
```
hdevalence: `Signature{ r, s }`

src/ed25519.rs
```diff
@@ -593,6 +593,91 @@ impl ExpandedSecretKey {
...
+        h.input(b"SigEd25519 no Ed25519 collisions");
+        h.input(&[1]); // Ed25519ph
+        h.input(&[ctx_len]);
+        h.input(ctx);
+        h.input(r.as_bytes());
+        h.input(public_key.as_bytes());
+        h.input(&prehash);
+        hash.copy_from_slice(h.fixed_result().as_slice());
+
+        hram_digest = Scalar::from_bytes_mod_order_wide(&hash);
```
hdevalence: ditto on `Scalar::from_hash`

src/ed25519.rs
```diff
@@ -758,6 +843,63 @@ impl PublicKey {
...
+        h.input(signature.r.as_bytes());
+        h.input(self.as_bytes());
+        h.input(prehashed_message.fixed_result().as_slice());
+        hash.copy_from_slice(h.fixed_result().as_slice());
+
+        let digest_reduced: Scalar = Scalar::from_bytes_mod_order_wide(&hash);
+        let r: EdwardsPoint = EdwardsPoint::vartime_double_scalar_mul_basepoint(&digest_reduced,
+                                                                                &a, &signature.s);
+
+        (signature.r.as_bytes()).ct_eq(r.compress().as_bytes()).unwrap_u8() == 1
```
hdevalence: This can just be `r.compress() == signature.r` (no CT needed)

src/ed25519.rs
```diff
@@ -758,6 +843,63 @@ impl PublicKey {
...
+        h.input(b"SigEd25519 no Ed25519 collisions");
+        h.input(&[1]); // Ed25519ph
+        h.input(&[ctx_len]);
+        h.input(ctx);
+        h.input(signature.r.as_bytes());
+        h.input(self.as_bytes());
+        h.input(prehashed_message.fixed_result().as_slice());
+        hash.copy_from_slice(h.fixed_result().as_slice());
+
+        let digest_reduced: Scalar = Scalar::from_bytes_mod_order_wide(&hash);
```
hdevalence: `Scalar::from_hash`

src/ed25519.rs
isislovecruft: My vague reasoning was that I wanted to force people to actually digest things with an appropriately sized hash, versus all the other things they could do if it just took a `[u8; 64]`.

isislovecruft: *waves arms in a squiggly fashion* ~~ T Y P E   S A F E T Y~~

src/ed25519.rs
isislovecruft: Yeah! I forgot we added that but saw it in your #30 code. I was going to change that and also change the `verify_prehashed()` method to return a `Result`.

src/ed25519.rs
isislovecruft: you kids n ur new fangled rust syntax, get offa my lawn

src/ed25519.rs
isislovecruft: Yep! I was going to change it to match your changes from #30 after the latter gets merged.

PR #32, opened 2018-07-15
Cleanup signing code to use new dalek APIs and Rust syntax.

isislovecruft:


src/ed25519.rs
```diff
@@ -110,7 +111,7 @@ impl PartialEq for Signature {
         let mut equal: u8 = 0;
 
         for i in 0..32 {
-            equal |= self.r.0[i] ^ other.r.0[i];
+            equal |= self.R.0[i] ^ other.R.0[i];
             equal |= self.s[i]   ^ other.s[i];
```
hdevalence: hmm, this isn't related to this change, but why doesn't `Signature` just `#[derive(Eq, PartialEq)]`?

src/ed25519.rs
```diff
@@ -562,34 +563,30 @@ impl ExpandedSecretKey {
...
 
         h.input(&self.nonce);
         h.input(&message);
-        hash.copy_from_slice(h.fixed_result().as_slice());
-
-        mesg_digest = Scalar::from_bytes_mod_order_wide(&hash);
 
-        r = (&mesg_digest * &constants::ED25519_BASEPOINT_TABLE).compress();
+        r = Scalar::from_hash(h);
+        R = (&r * &constants::ED25519_BASEPOINT_TABLE).compress();
```
hdevalence: I think it would be cleaner to do `let R = ...` here instead of defining the variable earlier

src/ed25519.rs
```diff
@@ -562,34 +563,30 @@ impl ExpandedSecretKey {
...
 
         h = D::default();
```
hdevalence: I think this should be `let mut h = ` since it's a new variable.

PR #33, opened 2018-07-20
Batch verification

isislovecruft:
See the commit message on ce46a12. There might be a couple ways to make this better, but imho this is working and can be merged now with further backend optimisations later.


hdevalence:
I have an extra patch on this branch here: https://github.com/isislovecruft/ed25519-dalek/compare/feature/27-batch...hdevalence:feature/27-batch that avoids doing any extra allocations, using new API in `curve25519-dalek`.

I think it'd be best to wait until we do a `1.0.0-pre.0` and then update this branch to avoid allocations.

PR #34, opened 2018-07-20
implement zero on drop for secret key material

isislovecruft:


hdevalence:
It would probably be good to document somewhere that the `Default` impl is there in order to have clear-on-drop-able stuff, but not for other use. Otherwise this looks good, I guess it's waiting on the next `curve25519-dalek` release.

PR #40, opened 2018-09-11
Add constructors for data that only validates size

coltfred:
Added constructors for the data types which only validate size in their `from_bytes`. This allows consumers who have already statically proven the size to avoid `.unwrap` calls that can never actually throw.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -185,6 +185,12 @@ impl Drop for SecretKey {
 }
 
 impl SecretKey {
+
+    ///Construct a SecretKey given an array of `SECRET_KEY_LENGTH`.
+    pub fn new(bytes: [u8; SECRET_KEY_LENGTH]) -> SecretKey {
+        SecretKey(bytes)
+    }
+
     /// Expand this `SecretKey` into an `ExpandedSecretKey`.
     pub fn expand<D>(&self) -> ExpandedSecretKey where D: Digest<OutputSize = U64> + Default {
         ExpandedSecretKey::from_secret_key::<D>(&self)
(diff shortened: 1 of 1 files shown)


coltfred:
@isislovecruft Would you be willing to give me some feedback on this PR?

coltfred:
Ping! I'd love to have some feedback on this PR if at all possible.

src/ed25519.rs
```diff
@@ -427,6 +433,19 @@ impl<'a> From<&'a SecretKey> for ExpandedSecretKey {
...
+
+    ///Construct an ExpandedSecretKey given an array of exactly `EXPANDED_SECRET_KEY_LENGTH` bytes.
+    fn new(bytes: [u8; EXPANDED_SECRET_KEY_LENGTH]) -> ExpandedSecretKey {
+        let mut lower: [u8; 32] = [0u8; 32];
+        let mut upper: [u8; 32] = [0u8; 32];
+
+        lower.copy_from_slice(&bytes[00..32]);
+        upper.copy_from_slice(&bytes[32..64]);
+
+        ExpandedSecretKey{ key:   Scalar::from_bits(lower),
```
isislovecruft: I don't trust my users to always hand me canonical bits in reduced form, and I don't want to provide them with constructors which they can use as foot cannons.

coltfred: I must be confused. https://github.com/dalek-cryptography/ed25519-dalek/blob/master/src/secret.rs#L401 provides a way to do exactly this code, but verifies the length. If we statically guarantee the length then the function is the same as `from_bytes` but without the failure condition. I can understand if you don't want to call it `new`. `from_sized_bytes` might be a better option.

Issue #39, opened 2018-09-11
Being able to compute PublicKey from an ExpandedSecretKey


coltfred:
I'm in a system where the user has an ExpandedSecretKey and I want to sign a message. I don't have access to the SecretKey byes that were used to produce the ExpandedSecretKey. Is there a way to compute the PublicKey without manually multiplying times the base point? Also I'm confused about why the `sign` function takes the PublicKey when it can be computed.

Sorry if I've missed something.

ValarDragon:
Ed25519 signing requires the public key. While it can be computed, that is an extra constant time scalar multiplication, which you could avoid. I believe the choice of having sign take in the pubkey was in acknowledgement of performance. Most of the time you would already have the pubkey before signing, thus theres little sense in recomputing it.

coltfred:
@ValarDragon In the case of the ExpandedSecretKey it's ugly to compute the key because the scalar is hidden as private to the crate and there isn't a `compute_public_key` on it. I don't care if I need to pass it, but having to manually multiply the base point, serialize to bytes and calling Scalar::from_bits is a bit much.

For clarity, here is my code:

```rust
        let private_key_bytes: [u8; 64] = //Some real value
        //This unwrap cannot fail. The only thing that the `from_bytes` does for validation is that the
        //value is 64 bytes long, which we guarentee statically.
        let secret_key: ExpandedSecretKey =
            ExpandedSecretKey::from_bytes(&private_key_bytes).unwrap();
        let secret_bytes: [u8; 32] = {
            let mut secret_bytes = [0; 32];
            secret_bytes.copy_from_slice(&private_key_bytes[0..32]);
            secret_bytes
        };
        let public_key_point =
            (&Scalar::from_bits(secret_bytes) * &constants::ED25519_BASEPOINT_TABLE).compress();
        let sig = secret_key.sign::<Sha512>(
            &message,
            &PublicKey::from_bytes(&public_key_point.to_bytes()).unwrap(),
        );
```

coltfred:
@isislovecruft @ValarDragon  Did you have any thoughts on this? I'd be happy to PR this change if you'd be willing to accept it. I'm thinking adding this to `ExpandedSecretKey`:
```rust
fn compute_public_key(&self) -> PublicKey 
```

isislovecruft:
[...]

PR #41, opened 2018-09-12
Implement From<ExpandedSecretKey> for PublicKey.

isislovecruft:
* CLOSES https://github.com/dalek-cryptography/ed25519-dalek/issues/39

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -777,22 +777,36 @@ impl PublicKey {
     /// Derive this public key from its corresponding `SecretKey`.
     #[allow(unused_assignments)]
     pub fn from_secret<D>(secret_key: &SecretKey) -> PublicKey
-            where D: Digest<OutputSize = U64> + Default {
-
+        where D: Digest<OutputSize = U64> + Default
+    {
         let mut h:    D = D::default();
         let mut hash:   [u8; 64] = [0u8; 64];
         let mut digest: [u8; 32] = [0u8; 32];
-        let     pk:     [u8; 32];
 
         h.input(secret_key.as_bytes());
         hash.copy_from_slice(h.fixed_result().as_slice());
 
         digest.copy_from_slice(&hash[..32]);
-        digest[0]  &= 248;
-        digest[31] &= 127;
-        digest[31] |= 64;
 
-        pk = (&Scalar::from_bits(digest) * &constants::ED25519_BASEPOINT_TABLE).compress().to_bytes();
+        PublicKey::mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(&mut digest)
+    }
+
+    /// Derive this public key from its corresponding `ExpandedSecretKey`.
+    pub fn from_expanded_secret(expanded_secret_key: &ExpandedSecretKey) -> PublicKey {
+        let mut bits: [u8; 32] = expanded_secret_key.key.to_bytes();
+
+        PublicKey::mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(&mut bits)
+    }
+
+    /// Internal utility function for mangling the bits of a (formerly
+    /// mathematically well-defined) "scalar" and multiplying it to produce a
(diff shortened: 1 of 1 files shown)


src/ed25519.rs
```diff
@@ -777,22 +777,37 @@ impl PublicKey {
...
+        let mut bits: [u8; 32] = expanded_secret_key.key.to_bytes();
+
+        PublicKey::mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(&mut bits)
+    }
+
+    /// Internal utility function for mangling the bits of a (formerly
+    /// mathematically well-defined) "scalar" and multiplying it to produce a
+    /// public key.
+    #[inline]
+    fn mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(bits: &mut [u8; 32]) -> PublicKey {
```
tarcieri: 🤣

hdevalence: I think this would be better named as `from_scalar_bits`... it's mathematically well-defined, it's just treating the secret as an integer, not as a scalar (integer mod l), which is fine for the purposes of ed25519 (which does not do hkd or multisig or ...)

src/ed25519.rs
```diff
@@ -777,22 +777,37 @@ impl PublicKey {
...
-        digest[0]  &= 248;
-        digest[31] &= 127;
-        digest[31] |= 64;
 
-        pk = (&Scalar::from_bits(digest) * &constants::ED25519_BASEPOINT_TABLE).compress().to_bytes();
+        PublicKey::mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(&mut digest)
+    }
+
+    /// Derive this public key from its corresponding `ExpandedSecretKey`.
+    pub fn from_expanded_secret(expanded_secret_key: &ExpandedSecretKey) -> PublicKey {
```
hdevalence: Why have both this function and a `From` impl? It seems like it would be simpler just to have this code in the `From` impl and have people call `PublicKey::from(expanded_secret_key)`.

src/ed25519.rs
```diff
@@ -885,6 +900,12 @@ impl PublicKey {
     }
 }
 
+impl<'a> From<&'a ExpandedSecretKey> for PublicKey {
```
hdevalence: This should just be `impl From<ExpandedSecretKey> for PublicKey`; using a borrow messes up the derived `Into` implementation.

And the borrow doesn't provide any benefit: it will almost certainly be eliminated by inlining, and even if it weren't, the worst case is a 32-byte stack copy, which is irrelevant compared to the cost of the basepoint multiplication.

src/ed25519.rs
```diff
@@ -777,22 +777,37 @@ impl PublicKey {
...
+    pub fn from_expanded_secret(expanded_secret_key: &ExpandedSecretKey) -> PublicKey {
+        let mut bits: [u8; 32] = expanded_secret_key.key.to_bytes();
+
+        PublicKey::mangle_scalar_bits_and_multiply_by_basepoint_to_produce_public_key(&mut bits)
+    }
+
+    /// Internal utility function for mangling the bits of a (formerly
+    /// mathematically well-defined) "scalar" and multiplying it to produce a
+    /// public key.
+    #[inline]
```
hdevalence: It's not necessary to mark this as `#[inline]` since it's not externally accessible.

src/ed25519.rs
isislovecruft: Fixed.

src/ed25519.rs
isislovecruft: Fixed.

isislovecruft:
Hey I just realised that since I stopped using gnupg when my key finally expired the other day and I switched to @withoutboats' [bpb](https://github.com/withoutboats/bpb), that the little [verified] signature tags on my commits here are being made with ed25519-dalek. :)

withoutboats:
dalek is now self hosting

PR #47, opened 2018-10-24
Fix serde doc tests

garious:
And let latest rustfmt reorder imports.


isislovecruft:
Hi @garious! Thanks so much for catching this and fixing it!

PR #48, opened 2018-10-24
Add tests for upgraded generic_array

garious:
generic_array v0.12 no longer serializes GenericArray as a Vec,
which reduces the serialized size of PublicKey, Signature, and
SecretKey by 8 bytes.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -1629,7 +1629,10 @@ mod test {
     }
 
     #[cfg(all(test, feature = "serde"))]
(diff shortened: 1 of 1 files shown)


isislovecruft:
Hi @garious! I'm not sure we have anything using `generic_array` anymore? Was it that your code was using both `ed25519-dalek` and different versions of `generic_array`?

I'm happy to merge this regardless, as more tests are always good, and as these are likely types that will cross FFI boundaries in some applications it's good to know if their size under a given serialisation scheme changes.

PR #49, opened 2018-10-24
Add AsRef instances for PublicKey and SecretKey

garious:
This just makes it a little easier to migrate to this library from alternatives such as 'ring'.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -183,6 +183,12 @@ impl Drop for SecretKey {
     }
 }
 
+impl AsRef<[u8]> for SecretKey {
+    fn as_ref(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
+
 impl SecretKey {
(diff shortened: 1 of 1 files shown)


isislovecruft:
Hi @garious! Thanks for the patch!

isislovecruft:
Accidentally merged onto `master` rather than `develop`, fixing...

src/ed25519.rs
```diff
@@ -183,6 +183,12 @@ impl Drop for SecretKey {
     }
 }
 
+impl AsRef<[u8]> for SecretKey {
```
tarcieri: I'm not sure if I'm overly paranoid, but elsewhere I've avoided `impl`ing `AsRef` directly on types which wrap secrets like this, preferring explicit APIs for accessing secrets.

That said, this is the sort of thing that could probably benefit from some shared, reusable, and explicit traits for secret-keeping types, e.g. "like `AsRef<[u8]>`, but instead `ExposeSecret<[u8]>`".

PR #50, opened 2018-10-24
Expose CI failures

garious:


garious:
Moved this patch into #47, which now fixes CI and the doc tests in the same PR.

PR #51, opened 2018-10-24
Rename from_bytes to from_slice and add a from_bytes that constructs from an array

mehcode:
I have a need for efficient deconstruction and reconstruction of public and secret keys from their byte representation.

If we're not comfortable with this breaking change perhaps another name for the method (so `from_bytes` doesn't need a rename). We could also `pub` the inner portion of the types.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -247,6 +247,12 @@ impl SecretKey {
         Ok(SecretKey(bits))
     }
 
+    /// Convert a byte array to a `SecretKey`.
+    #[inline]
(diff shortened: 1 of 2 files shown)


coltfred:
IMO, this should also have a `from_bytes` on ExpandedSecretKey. I made an attempt at something like this before, but there was some concern about if this presented some danger to the usability of the library. 

See https://github.com/dalek-cryptography/ed25519-dalek/pull/40#discussion_r264918961

isislovecruft:
Thanks, but this would be a breaking change at this point. Also we're following the naming conventions that curve25519-dalek uses.

PR #54, opened 2018-11-06
Update curve25519-dalek dependency to 1.0.0-pre.0.

isislovecruft:


src/ed25519.rs
```diff
@@ -549,7 +551,7 @@ impl ExpandedSecretKey {
     /// # fn main() { }
     /// ```
     pub fn from_secret_key<D>(secret_key: &SecretKey) -> ExpandedSecretKey
-            where D: Digest<OutputSize = U64> + Default {
+            where D: Digest<OutputSize = U64> + FixedOutput + Default {
```
hdevalence: I don't think you need to add `FixedOutput` here, it's already part of the `Digest` trait.

hdevalence:
```suggestion
            where D: Digest<OutputSize = U64> + Default {
```

src/ed25519.rs
```diff
@@ -618,9 +620,9 @@ impl ExpandedSecretKey {
                              prehashed_message: D,
                              public_key: &PublicKey,
                              context: Option<&'static [u8]>) -> Signature
-        where D: Digest<OutputSize = U64> + Default
+        where D: Digest<OutputSize = U64> + FixedOutput + Default
```
hdevalence:
```suggestion
        where D: Digest<OutputSize = U64> + Default
```

PR #57, opened 2018-11-15
Rand 0.6 version bump

coltfred:
Update to rand 0.6. The only breaking change was that the doctests and tests needed to change the some `use` blocks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,7 @@ exclude = [ ".gitignore", "TESTVECTORS", "res/*" ]
 travis-ci = { repository = "dalek-cryptography/ed25519-dalek", branch = "master"}
 
 [dependencies.curve25519-dalek]
-git = "https://github.com/IronCoreLabs/curve25519-dalek.git"
-branch = "rand-0.6"
+version = "1.0.0-pre.1"
 default-features = false
 
 [dependencies.rand]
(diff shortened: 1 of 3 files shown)


coltfred:
I've updated to depend on the pre1 version of curve25519. If we could merge this and publish a pre1 of ed25519, I'd be happy to test it out.

Thanks for all your hard work.

coltfred:
Ping. We'd love to test out a pre1 version with rand 0.6.

coltfred:
@isislovecruft @hdevalence Is there any chance of a pre1 build with this included (to match the curve-25519 pre1 build). We're in a holding pattern waiting for matching builds and it's blocking us from being able to publish to crates. If there's any other work that needs to be done, I'm happy to help.

hdevalence:
Hey, I don't actually have the ability to publish updates to this crate (I'm not the maintainer), but there are now non-prerelease versions of `subtle` and `curve25519-dalek` on crates.io.

aep:
@isislovecruft would you be ok adding @hdevalence or even the github team as a crate maintainer?

isislovecruft:
Hey @aep! Yes, I'd be happy to add @hdevalence and @cathieyun as cargo maintainers, if they want to be.

isislovecruft:
Hi @coltfred! Thanks for the patch!

W.r.t. making an `ed25519-dalek-1.0.0-pre.1` crate, I can put something out tomorrow but there were additional changes I was planning to make (e.g. #61) so there'd be (at least) another update.

coltfred:
Even if there were another update, it'd help us get off branch deps which
would be helpful. Thanks!

PR #68, opened 2019-01-05
Use rand_core

newpavlov:
`rand` crate is still used for thread RNG, but I think in future it will get splitted into a separate crate as well. Also I've fixed some of doctests for `sha2` feature. (I think you should add it to CI)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,10 +19,13 @@ travis-ci = { repository = "dalek-cryptography/ed25519-dalek", branch = "master"
 version = "1.0.0-pre.0"
 default-features = false
 
-[dependencies.rand]
-version = "0.5"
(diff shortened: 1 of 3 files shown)


newpavlov:
Conflicts are too confusing, so I think it will be easier to redo this PR from scratch. :/

Also note that `rand` project has [plans](https://github.com/rust-random/rand/issues/715) for the next breaking version, which may result in a breaking change for `rand_core`.

newpavlov:
Current code uses `ThreadRng` in `verify_batch`, which [right now](https://github.com/rust-random/rand/issues/716) is only accessible via `rand`. I could replace it with `rand_os`, but performance will be a bit worse in result.

So I guess it will be better to postpone these changes until `rand v0.7` is released.

Issue #35, opened 2018-08-21
no_std depends on bincode


aep:
https://gist.github.com/jamesmunns/8d4de492b1b6eb56f078a1fe975aa758

aep:
cc @jamesmunns

jamesmunns:
Context: We are considering trying to use ed25519 in a microcontroller context.

It looks like the dependency on `bincode` has a hard `std` dependency: https://github.com/TyOverby/bincode/issues/189

jamesmunns:
It also seems to complain about `byteorder`, but from a first glance I can't understand why (you specify no default features):

```
james@archmac13r2 ➜  ed25519-dalek git:(master) cargo build -j1 --no-default-features --target thumbv7em-none-eabihf
   Compiling proc-macro2 v0.4.13
   Compiling typenum v1.10.0                                                                     
   Compiling cc v1.0.18                                                                          
   Compiling unicode-xid v0.1.0                                                                  
   Compiling rand_core v0.2.1                                                                    
   Compiling byteorder v1.2.4                                                                    
   Compiling subtle v0.7.0                                                                       
   Compiling failure_derive v0.1.2                                                               
   Compiling cfg-if v0.1.5                                                                       
   Compiling rustc-demangle v0.1.9                                                               
error[E0463]: can't find crate for `std`                               ] 11/41: byteorder        
  |
  = note: the `thumbv7em-none-eabihf` target may not be installed

error: aborting due to previous error

For more information about this error, try `rustc --explain E0463`.
error: Could not compile `byteorder`.  
```

isislovecruft:
Hi @aep and @jamesmunns! Thanks for reporting this. It appears to be a bug in [curve25519-dalek](https://github.com/dalek-cryptography/curve25519-dalek), which this crate depends upon. `curve25519-dalek` depends upon `byteorder` for both buildtime and runtime, and I was able to compile [byteorder](https://github.com/BurntSushi/byteorder) with `cargo build -j1 --no-default-features --target thumbv7em-none-eabihf`. I think it may have something to do with `curve25519-dalek`'s two-stage build process where cargo isn't passing either the correct flags or targets to both build stages? (cc/ @hdevalence)

isislovecruft:
[...]

PR #72, opened 2019-01-27
Fix rand dependency, deal with unusedness warnings

nickray:
This should fix the `no_std` builds, solving https://github.com/dalek-cryptography/ed25519-dalek/issues/35 apart from the [`dev-dependencies` issue](https://github.com/rust-lang/cargo/issues/4866).

Wasn't sure about `core::default::Default`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ default-features = false
 [dependencies.rand]
 version = "0.6"
 features = ["i128_support"]
+default-features = false
 
 [dependencies.serde]
 version = "^1.0"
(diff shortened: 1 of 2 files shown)


jamesmunns:
Hey @isislovecruft, would it be possible to merge this PR? It would be great to use this on a microcontroller!

pietgeursen:
This PR is rad and I'm looking forward to seeing it merged!

I've just got some code running on a micro (STM32F4) that uses this branch of `ed25519-dalek`. Great work peeps! :tada: :tada: :tada:

tarcieri:
Something it might be worth adding to this PR is Travis CI config which makes sure it builds for a non-`std` target, similar to this:

https://github.com/RustCrypto/signatures/blob/master/.travis.yml#L26

That way future `no_std` regressions can be detected.

PR #73, opened 2019-02-25
Add sha3_512 to cfg features

sunhuachuang:
Not sure it is needed, but I use sha3 to hash, so make a pr.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -23,7 +23,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "serde")]
 use serde::{Deserializer, Serializer};
 
-pub use sha2::Sha512;
+use super::Sha512;
 
 use curve25519_dalek::digest::generic_array::typenum::U64;
 pub use curve25519_dalek::digest::Digest;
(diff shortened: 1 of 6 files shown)


src/ed25519.rs
```diff
@@ -90,13 +90,13 @@ pub fn verify_batch(
...
     public_keys: &[PublicKey],
-) -> Result<(), SignatureError>
-{
-    const ASSERT_MESSAGE: &'static [u8] = b"The number of messages, signatures, and public keys must be equal.";
-    assert!(signatures.len()  == messages.len(),    ASSERT_MESSAGE);
-    assert!(signatures.len()  == public_keys.len(), ASSERT_MESSAGE);
-    assert!(public_keys.len() == messages.len(),    ASSERT_MESSAGE);
- 
+) -> Result<(), SignatureError> {
+    const ASSERT_MESSAGE: &'static [u8] =
```
isislovecruft: Mixing automated reformatting with commits makes the actual changes you're making very hard to review and is generally considered bad practice in open source development.

isislovecruft:
We don't implement Ed448 signatures, and as such, according to [RFC8032 §8.9](https://tools.ietf.org/html/rfc8032#section-8.9), we have no need for SHA-3/SHAKE256.

PR #78, opened 2019-03-21
Use provided `Digest` in `_prehashed` functions 

dpc:
Wasn't this the intention? Otherwise - what's the point of taking prehashed image a `Digest`. Just let to user pass the result as an array, and don't bother with the bounds.

Another thing is - I'm surprised that normal `verify` is not a special case of `verify_prehashed`. `verify_prehashed` adds some custom prefixed header, and does  hash of a hash instead of the hash of the message.

Seems like introducing special-cases and making interoperability harder.

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -361,7 +361,7 @@ impl Keypair {
         context: Option<&'static [u8]>,
     ) -> Signature
     where
-        D: Digest<OutputSize = U64>,
+        D: Digest<OutputSize = U64> + Default,
     {
         let expanded: ExpandedSecretKey = (&self.secret).into(); // xxx thanks i hate this
 
@@ -441,7 +441,7 @@ impl Keypair {
         signature: &Signature,
     ) -> Result<(), SignatureError>
     where
-        D: Digest<OutputSize = U64>,
+        D: Digest<OutputSize = U64> + Default,
     {
         self.public.verify_prehashed(prehashed_message, context, signature)
     }
--- a/src/public.rs
+++ b/src/public.rs
@@ -216,7 +216,7 @@ impl PublicKey {
         signature: &Signature,
     ) -> Result<(), SignatureError>
     where
-        D: Digest<OutputSize = U64> + Digest,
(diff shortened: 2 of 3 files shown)


tarcieri:
> Just let to user pass the result as an array, and don't bother with the bounds.

I think this is bad from a misuse resistance perspective. The security of Schnorr signatures (of which EdDSA/Ed25519 is effectively an instantiation) relies on the input passing through a random oracle (i.e. Schnorr signatures are an application of the [Fiat-Shamir heuristic](https://en.wikipedia.org/wiki/Fiat%E2%80%93Shamir_heuristic)).

Accepting raw digest outputs risks the digest output being provided by an attacker. In such scenarios, [an attacker can produce a signature which appears to verify without knowing the corresponding discrete log](https://twitter.com/pwuille/status/1063582706288586752).

dpc:
@tarcieri Oh, right. You could provide some long named escape hatch functions with **loud warnings** in the documentation, I guess. It's ultimately up to the user of this library to make sure the oracle is there, and it's not *that* hard not to misuse.

burdges:
There is considerable overhead in asking developers to make a choice among primitives like hash functions, so ideally you should [hide that choice even if you permit it](https://github.com/w3f/schnorrkel/blob/master/src/context.rs).  

Also, verify is the original ed25519, while the prehashed variants were added latter, so it's surprising that verify_prehash does not use verify!

tarcieri:
Now I remember that Ed25519ph/Ed25519ctx effectively "triple hashes" the input (by performing Ed25519-style double hashing on an input digest), and therefore my comments about Schnorr and ROM are irrelevant 😅Guess I have too much ECDSA on the brain.

All that said, I've adopted a `Digest`-based interface for the [DigestSigner](https://docs.rs/signature/latest/signature/trait.DigestSigner.html) trait  in the [signature](https://crates.io/crates/signature) crate for the above reasons, and think those sorts of APIs are a better convention for the above reasons, even if those concerns are irrelevant here.

Issue #69, opened 2019-01-06
Release 1.0.0


dpc:
Are we there yet? :)

coltfred:
:+1: I'm also super eager to get to 1.0.0.

newpavlov:
Personally I don't think that 1.0 should be released if there is non-1.0 dependencies which are part of public API, like `digest` crate which is pulled via `sha2`, `failure` and `clear_on_drop`. And I don't plan on doing 1.0 release for RustCrypto crates until stabilization of const generics (and ideally addition of [MSRV](https://github.com/rust-lang/rfcs/pull/2495) as well).

tarcieri:
@newpavlov I think I saw some PRs to remove the `sha2` bits from the public API? If not, they could be feature-gated under something which calls out the instability (e.g. `digest-preview`).

So long as it's not a part of the public API, I think it's fine for ed25519-dalek to use not-quite-1.0 gems behind the scenes, and even bump major versions of these in a point release.

hdevalence:
> releasing subtle v2.0 just after several months since v1.0 was unfortunate.

Indeed it is unfortunate, but in the end I don't think it's really a big deal. There's nothing *wrong* with the `1.0` version, it's just that the traits are not as flexible as they should be, and fixing it required bumping the major version number -- even though the actual upgrade is pretty painless.

It's fine for other people to want to wait for various language changes before declaring their APIs "done", but I don't think that that's our plan.

dpc:
I don't think `ed25519-dalek` is wide-spread public dependency shared between projects. You can do major release every 3 months, and from the perspective of the user, it is not a big of a deal. Major releases are typically a problem in crates that establish interfaces between other crates, especially ones that are API-heavy. `futures`, logging libraries, frameworks, traits crates (like `digest`, and I guess `subtle`) etc. With `ed25519-dalek` I just want to get some stuff signed.


Also, from my perspective it doesn't matter if it's `1.0` or `0.9`. I'm just looking for the `pre.0` to disappear and some certainty that things are OK.

coltfred:
I also don't care at all about major version bumps. I do want a version with rand 0.6 and curve-25519 1.0 as a dependency, though.

isislovecruft:
@dpc 

> Also, from my perspective it doesn't matter if it's 1.0 or 0.9. I'm just looking for the pre.0 to disappear and some certainty that things are OK.

[...]

PR #79, opened 2019-03-21
Implement plain custom digest functions

dpc:
Any chance this could be supported, at least temporarily, for backward-compatibility with how things previously worked?

### DIFF
--- a/src/ed25519.rs
+++ b/src/ed25519.rs
@@ -254,6 +254,21 @@ impl Keypair {
         Keypair{ public: pk, secret: sk }
     }
 
+    /// Generate keypair using custom `Digest`
+    pub fn generate_with_digest<R, D>(csprng: &mut R) -> Keypair
+    where
+        R: CryptoRng + Rng,
+        D: Digest<OutputSize = U64> + Default,
+    {
+        let sk: SecretKey = SecretKey::generate(csprng);
+        let pk = PublicKey::from_secret_with_digest::<D>(&sk);
+
+        Keypair {
+            public: pk,
+            secret: sk,
+        }
+    }
+
     /// Sign a message with this keypair's secret key.
     pub fn sign(&self, message: &[u8]) -> Signature {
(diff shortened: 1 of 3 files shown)


tarcieri:
This seems trivial to implement yourself atop `curve25519-dalek`

dpc:
I'd be happy to put the `Digest` stuff under a `feature` flag.

dpc:
Alternatively I'd be happy to help maintain some fork of `ed25519-dalek-ext` or something.I don't really feel comfortable doing even "trivial" stuff on my own.

burdges:
I'd think an ed25519-oops crate could be a fork of the older ed25519-dalek, so you could use both concurrently.  If you're only worried about expanded secret key generation from the secret key then you could maybe transition to storing the full expanded secret key.  If you need alternative hashing in signing then you could transition with a version number or trial verification probably.

dpc:
So custom digest feature is just going to get removed from `ed25519-dalek`? :/

dpc:
I think I might just take the hit, implement a way for my users to convert their old signed artifacts, and just change everything to be SHA-512-based. However - can anyone asure me that I won't need to do it ever again? I mean ... I was using pre.0 and `Digest` support got slashed. Now I'll use pre.1, and there might be pre.2, pre.3 :D

burdges:
I think hashing was changed to comply with the standard.

tarcieri:
"Ed25519", both in RFC8032 form and before, has always used SHA-512

dpc:
@tarcieri `ed25519` had functions to allow any hashing to be used under the hood, with `SHA-512` being a default, I believe. I wasn't aware of RFC8032, or the fact that ed25519 intention was to implement it. I still don't understand all the nuances here. I'd really appreciate if you could help me get answers to  https://github.com/dalek-cryptography/ed25519-dalek/issues/69#issuecomment-482783711 . @tarcieri

dpc:
I've got my answers, and looks like I don't need this stuff now.

PR #90, opened 2019-07-12
Update rand

vks:


hdevalence:
This crate should use the same version of `rand` as `curve25519-dalek`, which won't change for a while: https://github.com/dalek-cryptography/curve25519-dalek/pull/263

PR #93, opened 2019-07-29
Impl Hash, Ord and PartialOrd for PublicKey and Signature

madadam:
This PR adds implementations of `Ord`, `PartialOrd` and `Hash` to `PublicKey` and `Signature`. The motivation behind this change is to allow using `PublicKey` and `Signature` as keys in `HashMap` / `HashSet` / `BTreeMap` and `BTreeSet` and also to support deriving those traits on types which contain `PublicKey` and/or `Signature` members.

### DIFF
--- a/src/public.rs
+++ b/src/public.rs
@@ -9,7 +9,12 @@
 
 //! ed25519 public keys.
 
+use core::cmp::Ord;
+use core::cmp::Ordering;
+use core::cmp::PartialOrd;
 use core::fmt::Debug;
+use core::hash::Hash;
+use core::hash::Hasher;
 
 use curve25519_dalek::constants;
 use curve25519_dalek::digest::generic_array::typenum::U64;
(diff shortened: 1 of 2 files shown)


isislovecruft:
Hi @madadam! Thanks for the PR!

The reason these were not implemented was due to the malleability in `PublicKey`s (potential small torsion component) and `Signature`s (non-reduced scalar, and the same malleability via small torsion component in the public key and group element). Hence there is no way to really "compare" public keys or signatures. Essentially, anyone can take any `Signature` and produce seven more "valid" (under the same public key, or again, any variant of it, since it's also malleable) signatures over the same message by adding a point of small order to the `R` field and the public key.

ustulation:
@isislovecruft 

> Hence there is no way to really "compare" public keys ...

I can understand why maybe why an "ordering" comparison maybe doesn't make sense i.e, what does it conceptually mean to have PublicKey0 > or < PublicKey1 etc. If the user have a particular notion of ordering, they can implement it themselves.

However an equality comparison is also a comparison (after all in assembly it's merely compare-if-equal or jump-if-equal etc,). And that seems to be [implemented as here](https://github.com/dalek-cryptography/ed25519-dalek/blob/5870607aa770f8f08114fb8a8eae7e605b82b7c9/src/public.rs#L38) already.

With all that as context, I don't understand why `Hash` is not allowed ? It's (almost) fundamental that if you can compare for equality (which is by the way already in the crate before this PR) you might as well have it be hashable. That's how unordered containers like a HashMap/Set work - check for Hash and in case of collision do an equality check.

I don't (yet) see any reason why you don't allow the ability to be hashable but allow for equality comparison.

Will you accept a PR if it only has hash implemented (no ord/partial-ord) ?

Issue #97, opened 2019-10-03
provide a `verify_strict()` function


isislovecruft:
Many users want a `verify_strict()` function which would reject signatures contain torsion components in the `R` portion of the signature, in the signer's corresponding public key, or in the product of `s` component and the basepoint, as well as those whose `s` component is non-reduced/non-canonical. This isn't strictly according to [the RFC](https://tools.ietf.org/html/rfc8032) which states [in §5.1.7 (3)](https://tools.ietf.org/html/rfc8032#section-5.1.7):

> Check the group equation [8][S]B = [8]R + [8][k]A'.  It's
> sufficient, but not required, to instead check [S]B = R + [k]A'.

So, technically, by doing the full malleability checks for torsion components we're not RFC compatible, because the full checks themselves are not backwards compatible with the original definition of the ed25519 primitive, as well as earlier IETF standardised versions of the primitive. However, many modern users want this anyway, and this is quite sensible particularly for protocol designers lacking algorithmic agility but who desire added safety at the cost of backwards incompatibility.

Note that it is **my very strong opinion** that if you want this, **you should be using a signature scheme which uses [Ristretto](https://ristretto.group)**.

isislovecruft:
An idea was suggested to me via private communication that we could short circuit on the `s < \ell` scalar check by checking against the encoded/canonicalised version, e.g.:

```rust
for i in 0..32 {
    if s_bytes[i] >= l_bytes[i] {
        return Err();
    }
}
```

However, this does not work.

First, `\ell = 2^252 + 27742317777372353535851937790883648493` in big-endian binary is `10110111_11001011_10101111_00111010_01011000_11000110_01001000_00011010_01101011_00111001_11101111_01000101_01111011_10011111_01111011_00101000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000000_00001000`. It is easy to see for this that taking the first byte `10110111`, all possible bytes which are less than to the first could still represent a big-endian number less than `\ell`. Similarly for the second byte, and so on. For the last byte, it is easy to see that (big-endian, again) `00010000 < 00001000 == true` and that `00001000 < 00001000 == false`. However, obviously there are numbers less than `\ell` whose last byte is `00001000`. (This is the reason historically that most libraries "check" `s < \ell` by doing the following hacky trick to ensure the most significant three bits are unset:

```rust
[...]

PR #98, opened 2019-10-07
add `verify_strict()` functionality

isislovecruft:
This provides a feature to enable/disable various levels of checking for scalar malleability, with the full reduction check as the default. It also provides a new `PublicKey.verify_strict()` function with checks for small torsion components in the public key and `R` portion of the signature.


isislovecruft:
This closes #97 btw.

PR #99, opened 2019-10-11
Optimisation for signature verification to succeed fast in checking the scalar

isislovecruft:
This is only done during signature verification.


elichai:
Pretty damn awesome :)
Someone should write a book full of those tricks lol

PR #101, opened 2019-10-21
Update to latest rand

mikelodder7:
This updates to rand 0.7.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,18 +19,9 @@ travis-ci = { repository = "dalek-cryptography/ed25519-dalek", branch = "master"
 version = "1"
 default-features = false
 
-[dependencies.rand_core]
-version = "0.3"
-default-features = false
-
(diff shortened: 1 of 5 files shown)


isislovecruft:
Hi @mikelodder7! Thank for the PR. This looks good to me. The only thing is that I would not want to cut a `ed25519-dalek-1.0.0` with this until `curve25519-dalek-2.0.0` (which will update to `rand-0.7`) is out, but I'm happy to cut more prereleases beforehand.

mikelodder7:
Yes if you could cut a prelease to 1.0.0-pre3 or something like that it would be great

exrook:
Now that `curve25519-dalek`2.0.0 is out could we see a `pre3` with the updated dependencies? I'm running into some annoyances from the `rand_core` mismatch with the latest release of `x25519-dalek` (now also on `curve25519-dalek 2.0.0`)

PR #106, opened 2019-12-09
Switch to using zeroize rather than clear_on_drop.

isislovecruft:


src/secret.rs
```diff
@@ -554,3 +547,23 @@ impl<'d> Deserialize<'d> for ExpandedSecretKey {
...
+
+        { // scope for the secret to ensure it's been dropped
+            let secret = SecretKey::from_bytes(&[0x15u8; 32][..]).unwrap();
+
+            secret_ptr = secret.0.as_ptr();
+        }
+
+        let memory: &[u8] = unsafe { ::std::slice::from_raw_parts(secret_ptr, 32) };
+
+        assert!(!memory.contains(&0x15));
```
tarcieri: I'm curious to know if there's any way to write these sort of tests around drop handlers doing things like clearing memory in a safer manner.

PR #107, opened 2019-12-10
Use `default-features = false` with `serde`

cbeck88:
It doesn't appear to me that ed25519-dalek crate needs any of the std-related
features of serde. But it turns them on anyways because it doesn't put
`default-features = false`.

This breaks no_std builds. Otherwise I think we could use 1.0.0-pre3 in mobilecoin.
I'm going to test this revision in our build and see if I'm right. I don't think
this is a breaking change from dalek's point of view.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ curve25519-dalek = { version = "2", default-features = false }
 merlin = { version = "1", default-features = false, optional = true, git = "https://github.com/isislovecruft/merlin", branch = "develop" }
 rand = { version = "0.7", default-features = false, optional = true }
 rand_core = { version = "0.5", default-features = false, optional = true }
-serde = { version = "1.0", optional = true }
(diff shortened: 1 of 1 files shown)


isislovecruft:
Thanks @cbeck88! Perhaps we should add also "serde/std" to the "std" feature? I can add that now.

tarcieri:
If `ed25519-dalek` doesn't need `serde/std`, it's probably better to leave it disabled. If others want it, they can still get to it through feature unification.

isislovecruft:
Feature unification? Neat, is that like doing "ed25519-dalek/serde/std"?

tarcieri:
The features that end up getting activated in a crate are the set union of all of the features activated by crates which use it as a dependency, so if someone wants access to `serde/std` they can still get to it by including themselves directly (which they'll almost certainly want to do anyway in order to actually make use of `serde`)

Issue #108, opened 2019-12-11
Build fails if `rand` feature is disabled


tarcieri:
```
   Compiling ed25519-dalek v1.0.0-pre.3
error[E0432]: unresolved import `rand`
  --> /Users/bascule/.cargo/registry/src/github.com-1ecc6299db9ec823/ed25519-dalek-1.0.0-pre.3/src/ed25519.rs:14:5
   |
14 | use rand::{CryptoRng, RngCore};
   |     ^^^^ use of undeclared type or module `rand`
```

[That line](https://github.com/dalek-cryptography/ed25519-dalek/blob/ee67f36/src/ed25519.rs#L14) needs a `#[cfg(feature = "rand")]` above it.

isislovecruft:
Ah, oops, thanks @tarcieri!

isislovecruft:
Fixed in 8a2e9af

PR #110, opened 2020-01-07
Fixup rand

garbageslam:
I tried to uprev our version of ed25519-dalek but I ran into two issues:

1: missing `#[cfg(feature = rand)]` in `secret.rs` ( I think this is a further case of #108 ), we are building with these features:

```
..., default-features = false, features = ["alloc", "nightly", "serde", "u64_backend"] }
```

2: we lost `Default` trait on `SecretKey`, I don't know if that was intentional API change? but atm we are using that anyways, please lmk! thank you

### DIFF
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -19,6 +19,7 @@ use curve25519_dalek::digest::Digest;
 use curve25519_dalek::edwards::CompressedEdwardsY;
 use curve25519_dalek::scalar::Scalar;
 
+#[cfg(feature = "rand")]
 use rand::{CryptoRng, RngCore};
 
 use sha2::Sha512;


tarcieri:
> I don't know if that was intentional API change?

IMO `SecretKey` should *not* have a `Default`

cbeck88:
@tarcieri yah I guess I agree -- but if it's going to be removed, I think there should be semantic versioning around the change

garbageslam:
regarding the "Default" thing: 
at this point we've been patching for like 4 months -- it will have little impact on mobilecoin how you decide to think about semver here. i'm happy to remove the Default thing if that will help the PR land

isislovecruft:
I removed `Default` on `SecretKey` on purpose. It was there because `clear_on_drop` required it, but with it, I would worry about mistakes in accidentally initialising a `SecretKey` to all zeros. I'm happy to take the rest of this PR.

garbageslam:
not really, we were just using it, we can stop using it

garbageslam:
rebased on master and removed the re-addition of Default

garbageslam:
hmm, after rebase, I again can't build at this revision due to std/no_std issues :(

Will investigate further

garbageslam:
closing in favor of https://github.com/dalek-cryptography/ed25519-dalek/pull/139#issue-462944259

PR #123, opened 2020-03-05
Bump criterion version

huitseeker:
Note this allows the use of benchmark comparison tooling such as [critcmp](https://github.com/BurntSushi/critcmp).


huitseeker:
Oops, thats not necessary.

PR #124, opened 2020-03-17
Use `ed25519` + `signature` interop crates

tarcieri:
(take 2: my first attempt was #102)

The `signature` crate provides `Signer` and `Verifier` traits generic over signature types:

https://github.com/RustCrypto/traits/tree/master/signature

There's presently an open call to stabilize the parts of its API needed by Ed25519 signatures and release a 1.0 version:

https://github.com/RustCrypto/traits/issues/78

The `ed25519` crate, based on the `signature` crate, provides an `ed25519::Signature` type which can be shared across multiple Ed25519 crates (e.g. it is also used by the `yubihsm` crate):

https://github.com/RustCrypto/signatures/tree/master/ed25519

This commit integrates the `ed25519::Signature` type, and changes the existing `sign` and `verify` methods (where applicable) to use the `Signer` and `Verifier` traits from the `signature` crate. Additionally, it replaces `SignatureError` with the `signature` crate's error type.

This has the drawback of requiring the `Signer` and/or `Verifier` traits are in scope in order to create and/or verify signatures, but with the benefit of supporting interoperability with other Ed25519 crates which also make use of these traits.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,10 +23,11 @@ features = ["nightly", "batch"]
 
 [dependencies]
 curve25519-dalek = { version = "2", default-features = false }
+ed25519 = { version = "1", default-features = false }
 merlin = { version = "1", default-features = false, optional = true, git = "https://github.com/isislovecruft/merlin", branch = "develop" }
 rand = { version = "0.7", default-features = false, optional = true }
 rand_core = { version = "0.5", default-features = false, optional = true }
-serde = { version = "1.0", default-features = false, optional = true }
+serde_crate = { package = "serde", version = "1.0", default-features = false, optional = true }
 sha2 = { version = "0.8", default-features = false }
 zeroize = { version = "1", default-features = false, features = ["zeroize_derive"] }
 
@@ -45,9 +46,10 @@ harness = false
 
 [features]
 default = ["std", "u64_backend"]
-std = ["curve25519-dalek/std", "serde/std", "sha2/std", "rand/std"]
+std = ["curve25519-dalek/std", "ed25519/std", "serde_crate/std", "sha2/std", "rand/std"]
 alloc = ["curve25519-dalek/alloc", "rand/alloc", "zeroize/alloc"]
 nightly = ["curve25519-dalek/nightly", "rand/nightly"]
+serde = ["serde_crate", "ed25519/serde"]
 batch = ["merlin", "rand"]
 # This feature enables deterministic batch verification.
 batch_deterministic = ["merlin", "rand", "rand_core"]
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -9,8 +9,6 @@
 
 //! ed25519 keypairs.
 
-use core::default::Default;
-
 #[cfg(feature = "rand")]
 use rand::{CryptoRng, RngCore};
 
@@ -26,13 +24,12 @@ pub use sha2::Sha512;
 use curve25519_dalek::digest::generic_array::typenum::U64;
 pub use curve25519_dalek::digest::Digest;
 
-#[cfg(all(feature = "batch", any(feature = "std", feature = "alloc")))]
-pub use crate::batch::*;
-pub use crate::constants::*;
-pub use crate::errors::*;
-pub use crate::public::*;
-pub use crate::secret::*;
-pub use crate::signature::*;
+use ed25519::signature::{Signer, Verifier};
+
+use crate::constants::*;
+use crate::errors::*;
+use crate::public::*;
+use crate::secret::*;
 
 /// An ed25519 keypair.
 #[derive(Debug)]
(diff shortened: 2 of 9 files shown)


tarcieri:
This PR is hopefully less invasive than #102, and retains the original `Signature` type as `InternalSignature` (as opposed to #102, which added an extension trait), and also retains the `InternalError` type.

I'll leave some line notes about the specific changes.

src/keypair.rs
```diff
@@ -241,20 +231,20 @@ impl Keypair {
         &self,
         prehashed_message: D,
         context: Option<&[u8]>,
-    ) -> Signature
+    ) -> ed25519::Signature
```
tarcieri: This is now returning the `Signature` type from the `ed25519` crate (which is re-exported as `ed25519_dalek::Signature`).

I wasn't sure if this was helpful or not to signal this is the `ed25519` crate's signal type versus an `InternalSignature`.

src/keypair.rs
```diff
@@ -9,8 +9,6 @@
 
 //! ed25519 keypairs.
```
tarcieri: I renamed this whole module from `ed25519` -> `keypair` so I was able to re-export the entire `ed25519` crate as `ed25519_dalek::ed25519`.

It only contains the `Keypair` type anyway, so hopefully that makes sense.

src/lib.rs
```diff
@@ -237,6 +239,8 @@
 #[macro_use]
 extern crate std;
 
+pub extern crate ed25519;
```
tarcieri: Note: original `ed25519` module renamed to `keypair`, the `ed25519` crate re-exported in its place, and the re-exports for all of the other modules (which were previously in the `ed25519` module) moved below.

src/lib.rs
```diff
@@ -245,20 +249,29 @@ extern crate merlin;
...
-pub use crate::ed25519::*;
+pub use curve25519_dalek::digest::Digest;
+
 #[cfg(all(any(feature = "batch", feature = "batch_deterministic"), any(feature = "std", feature = "alloc")))]
 pub use crate::batch::*;
+pub use crate::constants::*;
+pub use crate::errors::*;
+pub use crate::keypair::*;
+pub use crate::public::*;
+pub use crate::secret::*;
```
tarcieri: These were all previously located in the `ed25519` module.

src/signature.rs
```diff
@@ -190,45 +183,23 @@ impl Signature {
...
-        Ok(Signature {
+        Ok(InternalSignature {
             R: CompressedEdwardsY(lower),
             s: s,
         })
     }
 }
 
-#[cfg(feature = "serde")]
-impl Serialize for Signature {
```
tarcieri: Note: the `ed25519` crate has its own `serde` impls, so these are no longer necessary

tests/ed25519.rs
```diff
@@ -281,7 +285,7 @@ mod serialisation {
     #[test]
     fn serialize_signature_size() {
         let signature: Signature = Signature::from_bytes(&SIGNATURE_BYTES).unwrap();
-        assert_eq!(serialized_size(&signature) as usize, 72); // These sizes are specific to bincode==1.0.1
+        assert_eq!(serialized_size(&signature) as usize, 64); // These sizes are specific to bincode==1.0.1
```
tarcieri: The `serde` serializers used by the `ed25519` crate are slightly different from the ones currently in `ed25519_dalek` .

Namely, they serialize a signature as a 64-element tuple, and in doing so avoid a length prefix when using formats like `bincode`, hence the signature size changed.

Cargo.toml
```diff
@@ -23,10 +23,11 @@ features = ["nightly", "batch"]
 
 [dependencies]
 curve25519-dalek = { version = "2", default-features = false }
+ed25519 = { version = "= 1.0.0-pre.4", default-features = false }
 merlin = { version = "1", default-features = false, optional = true, git = "https://github.com/isislovecruft/merlin", branch = "develop" }
 rand = { version = "0.7", default-features = false, optional = true }
 rand_core = { version = "0.5", default-features = false, optional = true }
-serde = { version = "1.0", default-features = false, optional = true }
+serde_crate = { package = "serde", version = "1.0", default-features = false, optional = true }
```
tarcieri: This little hack is needed so the `serde` cargo feature can activate `ed25519/serde` in order to access that crate's `Serialize`/`Deserialize` impls.

tests/ed25519.rs
```diff
@@ -24,6 +24,8 @@ use sha2::Sha512;
 
 #[cfg(test)]
 mod vectors {
+    use ed25519::signature::Signature as _;
```
tarcieri: This is needed to access the [`Signature::from_bytes`](https://docs.rs/signature/1.0.0-pre.5/signature/trait.Signature.html) method.

Elsewhere in this PR (e.g. in the docs) I've used `TryFrom`/`TryInto` to avoid this. Perhaps we should pick a single approach and standardize on it? (e.g. it seems nice if Keypair and PublicKey also had `TryFrom` impls)

tarcieri:
I've updated this PR to use the final 1.0 releases of the `ed25519` and `signature` crates:

- https://docs.rs/ed25519/1.0.1/ed25519/
- https://docs.rs/signature/1.0.1/signature/

tarcieri:
Awesome, thanks! 🎉

PR #139, opened 2020-08-04
Make `use rand::...` gated on `cfg(feature = "rand")`

garbageslam:
This is no longer actively breaking our no_std build, but I think
it's still technically a minor bug, and further case of issue #108

This was https://github.com/dalek-cryptography/ed25519-dalek/pull/110
but that PR was very old and there were additional issues then which are moot now

### DIFF
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -17,6 +17,7 @@ use curve25519_dalek::digest::Digest;
 use curve25519_dalek::edwards::CompressedEdwardsY;
 use curve25519_dalek::scalar::Scalar;
 
+#[cfg(feature = "rand")]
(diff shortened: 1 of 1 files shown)


isislovecruft (approved):
Hi @garbageslam. Thanks for the PR.

I'm happy to merge but this means that the `SecretKey::generate` function below won't work unless the "rand" feature is present, so it'll need to be feature gated as well. Additionally, I think the "rand" feature should be on by default, otherwise this breaks some backwards compatibility.

isislovecruft:
There's also a few other places where `use rand::*` occurs, in keypair.rs and batch.rs.

PR #143, opened 2020-09-15
Add ExpandedSecretKey::sign_vectored

tomaka:
In situations where the message you want to sign is split into multiple chunks, such as when you want to sign a message with a prefix, it feels wasteful to have to make the message contiguous just to be able to pass a slice.

This PR therefore proposes to add an `ExpandedSecretKey::sign_vectored` method that accepts a message as a list of chunks.

I've opted to only modify `ExpandedSecretKey` and not the `Signer` trait in order to not break the public API, and because this is kind of a niche feature.
Since this is a niche feature for people like me who like premature optimizations, I'd understand if this was not accepted.

### DIFF
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -9,7 +9,7 @@
 
 //! ed25519 secret key types.
 
-use core::fmt::Debug;
+use core::{fmt::Debug, iter};
 
 use curve25519_dalek::constants;
 use curve25519_dalek::digest::generic_array::typenum::U64;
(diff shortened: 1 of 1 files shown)


burdges:
There is a standard prehash variant for this purpose https://docs.rs/ed25519-dalek/1.0.0/ed25519_dalek/struct.Keypair.html#method.sign_prehashed 

There are minor concern about prehashed variants when used with low quality hash functions, like perhaps sha1, but those worries should not apply to sha2 within the foreseeable future.

isislovecruft:
Hi @tomaka, thanks for the suggestion. I'm going to mirror @burdges' advice that this is precisely one of the use cases for the ed25519_ph ("pre-hashed") algorithm.

PR #146, opened 2020-09-21
Fix serde implementation for serde_json

xu-cheng:
Based on discussion in #140.

We use the [serde_bytes](https://github.com/serde-rs/bytes) crate for serialization implementations, which simplifies codes and fixes issues for serde_json.

### DIFF
--- a/tests/ed25519.rs
+++ b/tests/ed25519.rs
@@ -230,10 +230,10 @@ struct Demo {
 mod serialisation {
     use super::*;
 
-    use self::bincode::{serialize, serialized_size, deserialize, Infinite};
-    use self::toml;
-
     use ed25519::signature::Signature as _;
+
+    // The size for bincode to serialize the length of a byte array.
+    static BINCODE_INT_LENGTH: usize = 8;
 
     static PUBLIC_KEY_BYTES: [u8; PUBLIC_KEY_LENGTH] = [
         130, 039, 155, 015, 062, 076, 188, 063,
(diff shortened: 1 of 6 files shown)


tarcieri:
Interesting. Perhaps we should also consider using this in the `ed25519` crate to replace the current serializers, if they are similarly problematic with CBOR:

https://github.com/RustCrypto/signatures/blob/76008e8cd4d81203b791ec9672addd77c1e294f9/ed25519/src/lib.rs#L282-L332

xu-cheng:
The CI error seems to be unrelated to this PR.

> Interesting. Perhaps we should also consider using this in the ed25519 crate to replace the current serializers, if they are similarly problematic with CBOR:

I don't know. Basically, it is a choice between to serialize as `&[u8]` or `[u8; LEN]`. In some formats like bincode, the latter is clearly better as it doesn't need to store the bytes length. In other formats like json, there is actually no difference. Then there comes with CBOR, which requires to encode the length for each element in the array. As such, using dynamic bytes array is actually better.

Currently there are two main issues:
* There are inconsistent in the serde implementations. The signature is using `[u8; LEN]`, while the other types are using `&[u8]`. It would be nice to have a consistent serde implementation between all types.
* If we change the serde implementation in either ways, it would be breaking changes. 

My suggestion would be to merge this PR first, as it fixes json without introducing breaking changes. The we can figure out what is the long term approach, which would definitely require a breaking change.

tests/ed25519.rs
```diff
@@ -230,11 +230,11 @@ struct Demo {
 mod serialisation {
     use super::*;
 
-    use self::bincode::{serialize, serialized_size, deserialize, Infinite};
-    use self::toml;
-
     use ed25519::signature::Signature as _;
 
+    // The size for bincode to serialize the lenght of a byte array.
```
isislovecruft, line 235: s/lenght/length/

isislovecruft (commented):
Hi @xu-cheng, thanks for the PRs. (And thanks @tarcieri for the input.)

Let me see if I understood this correctly:

1) We had inconsistent types in the various serde impls (`&[u8]` versus `[u8; N]`).
2) One of these types (which one???) is breaking the ability to interface with serde-json.
3) This API fixes the inconsistency in a completely backwards compatible way (i.e. it should be fine to merge for ed25519-dalek version 1.0.1).
4) Later we should decide if we want to use (de-)serialisers which are optimised towards either formats like CBOR which store the length or formats like JSON which do not.

Did I understand everything correctly?

FWIW I believe this is a breaking change by upgrading bincode, since the `Infinite` struct went away.

isislovecruft:
Also, yes, the CI failure was due to #145; it should be fixed now.

xu-cheng:
> One of these types (which one???) is breaking the ability to interface with serde-json.

It is `&[u8]`, which requires special handle during deserialization. Here we use `serde_bytes` crate to handle it automatically.

> FWIW I believe this is a breaking change by upgrading bincode, since the Infinite struct went away.

bincode is a develop dependency. So updating it should be fine.

> Did I understand everything correctly?

Yes.

So currently, `Signature`'s serde is using `[u8; N]` (https://github.com/RustCrypto/signatures/blob/76008e8cd4d81203b791ec9672addd77c1e294f9/ed25519/src/lib.rs#L305), while all other types (public/secret key and keypair)  are using `&[u8]` in serde.

* For json, `[u8; N]` and `&[u8]` yield the same result.
* For formats like bincode. `[u8; N]` is slightly more efficient as it avoids storing the extra length of the array.
* For CBOR, `&[u8]` is way more efficient. https://github.com/dalek-cryptography/ed25519-dalek/pull/140#issuecomment-694744544

isislovecruft:
@xu-cheng Thanks!

I'm not sure which way to go in the future, but this looks like a good start to at least get JSON working. Thanks so much for fixing this!

PR #147, opened 2020-09-22
batch verification malleability issue when used with fully deterministic nonce generation

isislovecruft:
Issue found by @real_or_random and @jonasnick.

Only if compiled with `--features "batch_deterministic"`, when `verify_batch()` was called, an adversary not in control of the secret key(s) could use a signature malleability bug to take any signature, `(s_i, R_i)`, and produce a signature `(s'_i, R_i)` where e.g. `s'_i = -z_i` which would erroneously pass batch verification when purely deterministic nonce generation was used, due to the sigma protocol transcript used to seed the CSPRNG not including the `s` components of the signatures in the context. The fix is simply to add the `s` components to the transcript.

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -43,19 +43,24 @@ use crate::public::PublicKey;
 use crate::signature::InternalSignature;
 
 trait BatchTranscript {
-    fn append_hrams(&mut self, hrams: &Vec<Scalar>);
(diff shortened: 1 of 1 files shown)


burdges:
Appears I managed to do this correctly: https://github.com/w3f/schnorrkel/blob/master/src/batch.rs#L146

isislovecruft:
@burges Yes, I checked, you're fine.

PR #149, opened 2020-09-22
fix serde in no_std

xu-cheng:
Oops, I accidentally broke no_std support in #146. Sorry.

/cc @isislovecruft


isislovecruft (approved):
Thanks @xu-cheng!

I'm going to move the "serde_bytes/alloc" feature to be dependent on our alloc feature.

Issue #130, opened 2020-06-04
Documentation for PublicKey::verify_strict() is misleading


abacabadabacaba:
The documentation for `PublicKey::verify_strict()` method states that it verifies a signature in a way that eliminates signature malleability. Unfortunately, this is not what it actually does.

The first point in the documentation talks about scalar malleability. As it says, both `verify` and `verify_strict` check that a scalar is encoded canonically. However, if `legacy_compatibility` feature is enabled, this check is loosened for both `verify` and `verify_strict`. A user can expect `verify_strict` to remain strict even if `legacy_compatibility` feature is used, but this is not what happens.

The second point, which talks about point malleability of *R*, is more misleading. First, being able to add a small-order component to *R* is not a source of signature malleability at all: because the value of *R* is hashed (together with the message and the public key), any change to *R*, even the addition of a torsion component, will invalidate the signature. So, anyone who is not in possession of a private key cannot change *R* while keeping the signature valid, just as they can't change the message. And anyone who *is* in possession of a private key can generate any number of distinct valid signatures of the same message by generating *r* randomly rather than deterministically.

By the way, the README file is also misleading in this regard. It states that "doing so allows for signature malleability, meaning that there may be two different and "valid" signatures with the same key for the same message, which is obviously incredibly dangerous", but a key holder can generate any number of such signatures that will pass even the most strict checks.

What the function *does* (and which is not mentioned in the documentation) is rejecting a signature if the key or *R* consist *exclusively* of a small-order torsion component. This is because they check those points using `is_small_order()` method. Is this intended? It looks like `is_torsion_free()` is more relevant here.

Also, `verify_strict()` doesn't check that the encoding of *R* is canonical. `verify()` does this implicitly, by computing the encoding of the expected value of *R* and comparing is with the encoding of *R* byte-wise. Unlike `verify()`, `verify_strict()` compares unpacked points, so a point that is encoded non-canonically may go undetected (but this doesn't lead to signature malleability, see above).

isislovecruft:
Hi @abacabadabacaba! Thanks for creating this issue.

> The first point in the documentation talks about scalar malleability. As it says, both verify and verify_strict check that a scalar is encoded canonically. However, if legacy_compatibility feature is enabled, this check is loosened for both verify and verify_strict. A user can expect verify_strict to remain strict even if legacy_compatibility feature is used, but this is not what happens.

I've clarified this in ad2e75f.

[...]

PR #153, opened 2020-10-14
[test-only] Add test showing non-repudiation for verify, verify_strict

huitseeker:
This PR is a follow-up of #98 (and may help #130), which aims to demonstrate the issue brought by small-order public keys. It shows an example of crafting a (public_key, signature) that verifies against two distinct messages using `verify`, but fails using `verify_strict`.
This has consequences on the possibility to repudiate a signed contract of blockchain transactions.

For more details, see:
https://eprint.iacr.org/2020/1244

Joint work with @kchalkias @valerini

### DIFF
--- a/tests/ed25519.rs
+++ b/tests/ed25519.rs
@@ -28,7 +28,10 @@ use sha2::Sha512;
 
 #[cfg(test)]
 mod vectors {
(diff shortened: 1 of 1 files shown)


kchalkias (approved):
Good idea! LGTM

isislovecruft (approved):
Thanks @huitseeker and @kchalkias!

PR #154, opened 2020-11-25
Speed up compilation by avoiding syn via zeroize_derive

spacejam:
Avoiding `syn` cuts release and debug compilation on my machine from ~6.5s to ~4.5s, saving the ~4000 daily users of this crate at least 2 hours each day assuming the average CPU is less speedy than the 9900k I tested this with.


stjepang:
On my machine, this PR in combination with https://github.com/wascc/nkeys/pull/6 halves compilation time for `nkeys` (from 12 sec down to 6 sec).

isislovecruft (approved):

isislovecruft:
Hi @spacejam! Thanks for the PR.

isislovecruft:
I had to re-merge this into main without making a release to keep the MSRV I wanted to target for adding CI via github actions.

PR #156, opened 2020-12-31
Update to rand 0.8 and rand-core 0.6

roblabla:
Bumps rand and rand-core to the latest version.

Note that because rand is a public dependency (it is exposed in Keypair::generate), this is a breaking change and should be released only with a major version bump.


roblabla:
Obsoleted by #160

PR #159, opened 2021-01-24
Re-export rand crate

toxeus:
The APIs in `ed25519_dalek` depend on the `rand` crate.
This can cause conflicts when an application needs to
provide different versions of `rand` to satisfy their
interfaces. By re-exporting the `rand` crate this issue
is eliminated.


toxeus:
ping @isislovecruft

PR #160, opened 2021-02-01
Update rand, rand-core (and merlin)

bltavares:
Updates rand and rand-core to their latest versions. It also requires updating merlin to their latest version to match the rand dependencies.

This PR contains commits from https://github.com/dalek-cryptography/ed25519-dalek/pull/156

Closes https://github.com/dalek-cryptography/ed25519-dalek/pull/156

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,8 @@ features = ["nightly", "batch"]
 curve25519-dalek = { version = "3", default-features = false }
 ed25519 = { version = "1", default-features = false }
 merlin = { version = "2", default-features = false, optional = true }
-rand = { version = "0.7", default-features = false, optional = true }
-rand_core = { version = "0.5", default-features = false, optional = true }
+rand = { version = "0.8", default-features = false, optional = true }
+rand_core = { version = "0.6", default-features = false, optional = true }
 serde_crate = { package = "serde", version = "1.0", default-features = false, optional = true }
 serde_bytes = { version = "0.11", optional = true }
 sha2 = { version = "0.9", default-features = false }
@@ -37,7 +37,7 @@ hex = "^0.4"
 bincode = "1.0"
 serde_json = "1.0"
 criterion = "0.3"
-rand = "0.7"
+rand = "0.8"
 serde_crate = { package = "serde", version = "1.0", features = ["derive"] }
 toml = { version = "0.5" }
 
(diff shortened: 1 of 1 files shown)


shekohex (approved):

robjtede:
A needed change but would be technically breaking unless a version range like `>=0.7, <0.9` is used.

toxeus:
This crate could easily prevent such breaking changes by re-exporting crates that are used by its public API such as `rand`.
I made a PR to fix this https://github.com/dalek-cryptography/ed25519-dalek/pull/159.

Well-maintained crates are doing this, e.g. [secp256k1](https://docs.rs/secp256k1/0.20.1/secp256k1/index.html#reexports).

roblabla:
> This crate could easily prevent such breaking changes by re-exporting crates that are used by its public API such as rand

It'd still be a breaking change though, as users could still use `rand` without going through the public re-export. It'd still be a good idea for usability reasons though.

toxeus:
Agreed, the user can always shoot herself in the foot. There's no silver bullet but it would be at least an improvement.

roblabla:
Bump. I'm currently keeping rand 0.7 around for dalek-cryptography, and it's causing me a bit of trouble, as the rest of the ecosystem is moving towards rand 0.8. Is there any way I can help moving this forward? `merlin` seems to have been updated (docs.rs lists latest version, 3.0.0, as using the correct version of rand). What are the remaining steps?

toxeus:
This project's maintenance seems to have dropped significantly. No activity by the maintainers on pull-requests and no commits since September 2020. I'm considering to maintain a fork and make sure that dependency updates land in a timely manner which is not really difficult using the auto-update tooling widely available (renovate-bot, dependa-bot, etc.). Let me know if there's interest. That might help my motivation :)

thomaseizinger:
> I'm considering to maintain a fork

This has already happened for the primary dependency of this crate: `curve25519-dalek`: https://github.com/zkcrypto/curve25519-dalek-ng

There is an open issue about forking `ed25519-dalek` into the same organization: https://github.com/zkcrypto/curve25519-dalek-ng/issues/8

toxeus:
I've migrated to https://github.com/jedisct1/rust-ed25519-compact. It's well maintained and works great.

rozbb:
All deps have been updated in the `release/2.0` branch. We will be cutting a prerelease soon.

PR #167, opened 2021-04-12
Remove rand from public api.

### LINKED ISSUE #162: `rand` update to v0.8
The package [`rand`](https://crates.io/crates/rand) has been updated to v0.8.X.
As this crate used rand v0.7.X it is not compatible with the current version (1.0.1) of this crate.

And will show error like this:
```
error[E0277]: the trait bound `OsRng: rand_core::CryptoRng` is not satisfied
   --> lib.rs:20:46
    |
315 |     let keypair: Keypair = Keypair::generate(&mut csprng);
    |                                              ^^^^^^^^^^^ the trait `rand_core::CryptoRng` is not implemented for `OsRng`
    | 
   ::: ed25519-dalek-1.0.1/src/keypair.rs:129:12
    |
129 |         R: CryptoRng + RngCore,
    |            --------- required by this bound in `Keypair::generate`

error[E0277]: the trait bound `OsRng: rand_core::RngCore` is not satisfied
   --> lib.rs:20:46
    |
315 |     let keypair: Keypair = Keypair::generate(&mut csprng);
    |                                              ^^^^^^^^^^^ the trait `rand_core::RngCore` is not implemented for `OsRng`
    | 
   ::: ed25519-dalek-1.0.1/src/keypair.rs:129:24
    |
129 |         R: CryptoRng + RngCore,
    |                        ------- required by this bound in `Keypair::generate`
```

On the code that looks like this (similar to example code in documentation).
```rust
use rand::rngs::OsRng;

let mut csprng = OsRng {};
let keypair: Keypair = Keypair::generate(&mut csprng);
```

So for the next version of this crate the `rand` version should be updated.
Current workaround is to downgrade to `rand` version 0.7.X.

dvc94ch:
I expect for most users this is fine, and if some user would like to use a different `Rng` they can use the `from_bytes` call.

Since `rand` has a history of causing breakage in `ed25519-dalek` I think this change could be justified as a bug fix.

Also updates the dependencies to new versions where appropriate.

Closes #160
Closes #159
Closes #162

### DIFF
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -10,7 +10,7 @@
 //! ed25519 keypairs.
 
 #[cfg(feature = "rand")]
-use rand::{CryptoRng, RngCore};
+use rand::{rngs::OsRng, CryptoRng, RngCore};
 
 #[cfg(feature = "serde")]
 use serde::de::Error as SerdeError;
@@ -95,6 +95,28 @@ impl Keypair {
     /// # Example
     ///
     /// ```
+    /// extern crate ed25519_dalek;
+    ///
+    /// # #[cfg(feature = "std")]
+    /// # fn main() {
+    /// #
+    /// use ed25519_dalek::Keypair;
+    /// let keypair = Keypair::generate();
+    /// # }
+    /// #
+    /// # #[cfg(not(feature = "std"))]
+    /// # fn main() { }
+    /// ```
+    #[cfg(feature = "rand")]
+    pub fn generate() -> Keypair {
+        Self::generate_with_rng(&mut OsRng {})
+    }
+
+    /// Generate an ed25519 keypair.
+    ///
+    /// # Example
+    ///
+    /// ```
     /// extern crate rand;
     /// extern crate ed25519_dalek;
     ///
@@ -106,7 +128,7 @@ impl Keypair {
     /// use ed25519_dalek::Signature;
     ///
     /// let mut csprng = OsRng{};
-    /// let keypair: Keypair = Keypair::generate(&mut csprng);
+    /// let keypair: Keypair = Keypair::generate_with_rng(&mut csprng);
     ///
     /// # }
     /// #
@@ -124,11 +146,11 @@ impl Keypair {
     /// which is available with `use sha2::Sha512` as in the example above.
     /// Other suitable hash functions include Keccak-512 and Blake2b-512.
     #[cfg(feature = "rand")]
-    pub fn generate<R>(csprng: &mut R) -> Keypair
+    pub fn generate_with_rng<R>(csprng: &mut R) -> Keypair
     where
         R: CryptoRng + RngCore,
     {
-        let sk: SecretKey = SecretKey::generate(csprng);
+        let sk: SecretKey = SecretKey::generate_with_rng(csprng);
         let pk: PublicKey = (&sk).into();
 
         Keypair{ public: pk, secret: sk }
@@ -154,18 +176,15 @@ impl Keypair {
     ///
     /// ```
     /// extern crate ed25519_dalek;
-    /// extern crate rand;
     ///
     /// use ed25519_dalek::Digest;
     /// use ed25519_dalek::Keypair;
     /// use ed25519_dalek::Sha512;
     /// use ed25519_dalek::Signature;
-    /// use rand::rngs::OsRng;
     ///
     /// # #[cfg(feature = "std")]
     /// # fn main() {
-    /// let mut csprng = OsRng{};
-    /// let keypair: Keypair = Keypair::generate(&mut csprng);
+    /// let keypair: Keypair = Keypair::generate();
     /// let message: &[u8] = b"All I want is to pet all of the dogs.";
     ///
     /// // Create a hash digest object which we'll feed the message into:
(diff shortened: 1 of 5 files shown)


src/keypair.rs
```diff
@@ -95,40 +92,24 @@ impl Keypair {
...
-    ///
-    /// The caller must also supply a hash function which implements the
-    /// `Digest` and `Default` traits, and which returns 512 bits of output.
-    /// The standard hash function used for most ed25519 libraries is SHA-512,
-    /// which is available with `use sha2::Sha512` as in the example above.
-    /// Other suitable hash functions include Keccak-512 and Blake2b-512.
     #[cfg(feature = "rand")]
-    pub fn generate<R>(csprng: &mut R) -> Keypair
-    where
-        R: CryptoRng + RngCore,
```
tarcieri, line 129: Note that this precludes use by embedded users who don't have an `OsRng` and instead have an owned RNG seeded by a hardware peripheral.

Ideally `rand` (or at least `rand_core`) will be stabilized and integrated into the language proper. IMO it would be good to check the `rust-rand` roadmap to see if that's what's planned for the next release. If there's only one more such upgrade, perhaps hiding it from the public API is unnecessary.

tarcieri:
> ...if some user would like to use a different Rng they can use the from_bytes call.

This has a corollary: if you'd like to avoid `rand`-related breakages, use `from_bytes`

dvc94ch:
Well, the downside is that many users would have to wrap all of the ed25519-dalek api to get a simple generate function. It also precludes me from being able to use ed25519-dalek in my public apis. Also I'd assume it is more advanced users that want to customize the rng, since zeroizing the original bytes can be a bit tricky. Every time the bytes are moved on the stack they need to be zeroized. But this is probably a problem even in ed25519-dalek (when you pass a keypair by value to a function) I guess?

dvc94ch:
Would adding a `generate_with_rng` method work in addition to a `generate` that just works? Although that is probably hard to sell as a bug fix :)

tarcieri:
> I'd assume it is more advanced users that want to customize the rng

It's needed by any user who is using a platform which isn't supported by `OsRng`:

https://docs.rs/getrandom/0.2.2/getrandom/#supported-targets

There is a new [`register_custom_getrandom`](https://docs.rs/getrandom/0.2.2/getrandom/macro.register_custom_getrandom.html) function in `getrandom`, but it is by design stateless and thus somewhat unhelpful for cases where a stateful, seeded RNG (or RNG peripheral) is needed.

dvc94ch:
Interesting, so maybe `generate_with_rng` should take a `FnMut(&mut [u8]) -> Result<(), E>` and `generate` would call `generate_with_rng(|dest: &mut [u8]| getrandom(dest))`. For key generation that's the paranoid thing to do anyway as opposed to seeding a csprng right?

dvc94ch:
So tried the `FnMut` closure approach. One added benefit is that we don't have to depend on rand for default builds at all anymore.

dvc94ch:
Are there any objections? Other than the marker trait for csprng I don't think we lose anything with this approach.

tarcieri:
This is definitely a breaking change and as such should be part of a major release.

It would also be good to look into the roadmap for `rand`/`rand_core` and whether they plan on putting out more breaking releases or if `rand` v0.8 / `rand_core` v0.6 are to be the last before `rand_core` is incorporated into `core`.

If the API is almost stable, I'm not sure it's worth dropping support now.

dvc94ch:
I didn't find a roadmap, could you point me to it?

> If the API is almost stable, I'm not sure it's worth dropping support now.

Well, it's not dropping support, they're still interoperable. `|bytes] OsRng {}.fill_bytes(bytes)` can hardly be considered "dropping support". It's just not clear what the advantage is of pulling in `rand` for `OsRng` which just calls out to `getrandom` and unwraps the result. In any case having a `generate` that doesn't require passing in a rng would be great.

tarcieri:
> It's just not clear what the advantage is of pulling in rand for OsRng which just calls out to getrandom and unwraps the result.

The use cases where `CryptoRng + RngCore` are interesting are ones where `OsRng` isn't being used, such as embedded devices supplying their own RNG, or using something like `rand::thread_rng()`.

I'm not sure the bounds you've supplied for `generate_with_rng` cover both of these cases and they can have conflicting requirements. For example, `thread_rng` is `Send + Sync`, but an embedded RNG might explicitly be `!Send`.

dvc94ch:
it's a FnMut so you can have arbitrary mutable state in it, and &mut FnMut also implements FnMut. And if you have something like this:
```rust
static mut RNG: HardwareRng = HardwareRng {};
```
you should be able to call `generate_with_rng(|bytes| RNG.fill_bytes(bytes))`

I think the Send/Sync concern doesn't apply as there is no bound for Send or Sync, meaning you can supply any FnMut not just a Send or a Sync one. I think `Sized` is applied automatically to generic arguments.

tarcieri:
You can't just stick hardware RNGs (or `SeedableRng`s seeded from them) in statics without resorting to convoluted mechanisms, and generally in the Rust Embedded ecosystem global state like that is frowned upon.

Hardware peripheral access in the Rust Embedded ecosystem is generally managed explicitly via owned values which are passed as parameters to functions after device initialization. This avoids any sort of racy "life before main"-style initialization problems.

tarcieri:
Let me put it this way:

- A convenience function which uses `OsRng` seems OK. You don't need to explicitly pull in `getrandom` though, or even `rand` to achieve that. If the `std` feature of `rand_core` is enabled, `rand_core::OsRng` will be available, wrapping `getrandom`
- I don't think it makes sense to define a `generate_with_rng` function that accepts a parameter other than something like `impl CryptoRng + RngCore`

dvc94ch:
third time is the charm

dvc94ch:
ping @isislovecruft

dvc94ch:
Closing for now as there doesn't seem to be any interest. Feel free to reopen if things change.

PR #180, opened 2021-09-14
WIP Updated deps

rozbb:
This depends on https://github.com/dalek-cryptography/curve25519-dalek/pull/369. The tests don't currently compile because `curve25519-dalek` expects the wrong version of `rand`. Also, **this requires a major version bump** for the same reason as the aforementioned PR.

Notes on the deps I updated:
* zeroize [obeys semantic versioning](https://github.com/iqlusioninc/crates/blob/main/zeroize/CHANGELOG.md). I left it at "1.4" (rather than "1") because newness here should be prioritized over deduplication
* bincode says their format is [stable across minor revisions](https://github.com/bincode-org/bincode/blob/trunk/readme.md)
* Merlin transcripts [have not changed since v2](https://github.com/zkcrypto/merlin/blob/main/CHANGELOG.md)
* Criterion requires the `html_reports` feature in order to not display an annoying warning

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,20 +24,20 @@ features = ["nightly", "batch"]
 [dependencies]
 curve25519-dalek = { version = "3", default-features = false }
 ed25519 = { version = "1", default-features = false }
-merlin = { version = "2", default-features = false, optional = true }
-rand = { version = "0.7", default-features = false, optional = true }
-rand_core = { version = "0.5", default-features = false, optional = true }
+merlin = { version = "3", default-features = false, optional = true }
+rand = { version = "0.8", default-features = false, optional = true }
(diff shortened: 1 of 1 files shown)


weikengchen:
@rozbb I found that curve25519-dalek's `4.0.0-pre.1` already has the new version of `rand`.

stephank:
I made these changes to get things to work with curve25519-dalek 4.0.0-pre.2: https://github.com/stephank/ed25519-dalek/commit/808b1e6f8a546345787261640c6c237f11ca0e19. CI passes.

It looks like the ed25519 crate has raised MSRV to 1.56 in a minor release, because it switched to Rust 2021. We could restrict its version to `>=1, <1.4`, or raise the MSRV here. I chose the latter because this is a change for a major release any way.

PR #184, opened 2021-12-06
Update curve25519-dalek's dependency of `rand`

weikengchen:
This PR is not to `main`, but to `update-deps`. 

This PR updates the dependency of `curve25519-dalek` to a version that uses `rand = 0.8`.

Since the code in `update-deps` is a breaking change, I would wait for a newer release of ed25519-dalek.


rozbb:
We currently pull in `curve25519-dalek` v4.0.0-pre.3 in the `release/2.0` branch

PR #196, opened 2022-04-23
Add `StreamVerifier`

robjtede:
## Context
In web applications, it is desirable to avoid storing the request payload. It is currently impossible to verify a signature based on a request payload that is receieved in a chunk stream without allocating additional memory equal to the size of the payload.

## Solution
A stateful type that holds the public key, candidate signature, and hash state that can be updated incrementally.

## To Do
- [x] decide if this type is appropriate for `ed25519-dalek`
- [x] type name
- [x] add `verify_stream` method to `Keypair`
- [x] documentation

## References
- As an example of the difference this feature would make: https://github.com/robjtede/actix-web-lab/pull/34/files

### DIFF
--- a/src/verifying/stream.rs
+++ b/src/verifying/stream.rs
@@ -1,0 +1,58 @@
+use curve25519_dalek::{edwards::EdwardsPoint, scalar::Scalar};
+use sha2::{Digest, Sha512};
+
+use crate::{signature::InternalSignature, InternalError, SignatureError, VerifyingKey};
+
+/// An IUF verifier for ed25519.
+///
+/// Created with [`VerifyingKey::verify_stream()`] or [`SigningKey::verify_stream()`].
+///
+/// [`SigningKey::verify_stream()`]: super::SigningKey::verify_stream()
+#[derive(Debug)]
+pub struct StreamVerifier {
+    /// Public key to verify with.
+    pub(crate) public_key: VerifyingKey,
+
+    /// Candidate signature to verify against.
+    pub(crate) signature: InternalSignature,
+
+    /// Hash state.
+    pub(crate) hasher: Sha512,
+}
+
+impl StreamVerifier {
+    /// Constructs new stream verifier.
+    ///
+    /// Seeds hash state with public key and signature components.
+    pub(crate) fn new(public_key: VerifyingKey, signature: InternalSignature) -> Self {
+        let mut hasher = Sha512::new();
+        hasher.update(signature.R.as_bytes());
+        hasher.update(public_key.as_bytes());
+
+        Self {
+            public_key,
+            hasher,
+            signature,
+        }
+    }
+
+    /// Digest message chunk.
+    pub fn update(&mut self, chunk: impl AsRef<[u8]>) {
+        self.hasher.update(&chunk);
+    }
+
+    /// Finalize verifier and check against candidate signature.
+    #[allow(non_snake_case)]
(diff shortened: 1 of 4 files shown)


tarcieri:
Note that there's a `DigestVerifier` trait (along with the `DigestSigner` trait) in the `signature` crate, which `ed25519-dalek` uses:

https://docs.rs/signature/latest/signature/trait.DigestVerifier.html

I think it would probably make sense for those to map to Ed25519ph though.

robjtede:
Yeah I saw that trait but it doesn't seem to map well to ed25519 since it receives the signature _after_ the message digest has been created.

tarcieri:
Ed25519ph is intended to be the IUF mode for Ed25519, and is already implemented by `ed25519-dalek` in the form of `sign_prehashed` and `verify_prehashed`.

Personally I would find it confusing to encounter an IUF mode for Ed25519 that wasn't Ed25519ph.

robjtede:
If ed25519 and ed25519ph verifications algorithms were compatible, I wouldn't have an issue.

What do you think is the best course of action for the use case listed of more-memory-efficient Discord webhook verification given I cannot change Discord itself to sign its requests using the *ph variant? This is clearly possible to acomplish unless the approach of this PR has security downsides I'm not aware of.

tarcieri:
You could potentially give it a different name which is less likely to be confused with Ed25519ph, like "StreamingVerifier"

robjtede:
Ahh. If this is just a naming issue then sure that seems very appropriate.

tarcieri:
Yeah, I think it might be nice for `ed25519-dalek` to eventually implement the `DigestSigner` and `DigestVerifier` traits from the `signature` crate for Ed25519ph, in which case it'd be nice to have a different name for an IUF verifier for traditional Ed25519

Monadic-Cat:
Heya! What's the status of this PR? It'd be nice to have for twilight-rs, so I'd like to help if there's anything I can do.

tarcieri:
Could you potentially use the `hazmat` low-level verifier API that was just added in #299?

Something like this might be nice as a safer API, but it should probably be (re)built in terms of the APIs introduced in #299.

Monadic-Cat:
> 

Sure! I'll try and do that today.

src/stream_verifier.rs
```diff
@@ -0,0 +1,58 @@
+use curve25519_dalek::{edwards::EdwardsPoint, scalar::Scalar};
```
tarcieri, line 1: Perhaps this module could be named `src/verifying/stream.rs` or `src/verifying/streaming.rs`?

src/verifying/stream.rs
```diff
@@ -0,0 +1,58 @@
...
+        self.hasher.update(&chunk);
+    }
+
+    /// Finalize verifier and check against candidate signature.
+    #[allow(non_snake_case)]
+    pub fn finalize_and_verify(self) -> Result<(), SignatureError> {
+        let minus_A: EdwardsPoint = -self.public_key.point;
+        let k = Scalar::from_hash(self.hasher);
+        let R =
+            EdwardsPoint::vartime_double_scalar_mul_basepoint(&k, &(minus_A), &self.signature.s);
```
tarcieri, line 50: I wonder if `VerifyingKey::recover_R` could be decomposed into a `VerifyingKey::recover_R_from_k` or thereabouts which captures this logic.

cc @rozbb

robjtede:
migrated to

Issue #121, opened 2020-02-25
Accept generic Digest


mgjm:
As explained in [#69 (comment)](https://github.com/dalek-cryptography/ed25519-dalek/issues/69#issuecomment-482871703):

> Ed25519 is a concrete instantiation of the EdDSA signature framework, which is generic over both an Edwards curve and a hash function (H). Ed25519 instantiates EdDSA with the “edwards25519” (RFC8032 terminology) elliptic curve and SHA-512.

But for some projects it would be useful to instantiates EdDSA with curve25519 and `Blake2b`.

Therefore I would suggest to split the `ed25519-dalek` crate into two crates:
- `eddsa-dalek`: Generic in terms of the `Digest` implementation
- `ed25519-dalek`: Hardcoded `sha2::Sha512` digest

I set up a test workspace with these two separate crates:
[mgjm/split-ed25519-dalek](https://github.com/mgjm/split-ed25519-dalek)

The `ed25519-dalek` crate only needs to reexport all non generic types and functions and export type aliases with hardcoded `Sha512` digest:
[`ed25519-dalek/src/lib.rs`](https://github.com/mgjm/split-ed25519-dalek/blob/32f756e/ed25519-dalek/src/lib.rs#L254-L264)

All tests and benchmarks using the `ed25519-dalek` crate still work without any modification.

---

This issue is also related to #75 and #79.

isislovecruft:
After internal discussion on the dalek team we decided that 1) the API complications combined with 2) making it easier for people to do EdDSA signatures in a non-spec-compliant manner meant we were not going to continue to support generic digests.

PR #201, opened 2022-06-01
Add generic methods to support non-sha512 digests

ryankurte:
apparently some folks like to use `ed25519` with other hashers (like `keccak` and `sha3`)... to use `dalek` in place of existing libraries exporting these functions one requires a mechanism for specifying the `digest` type.

this adds generic methods over a `Digest` for generating public keys and signing and verifying messages, and updates the existing API to call these to avoid duplication.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,3 +63,5 @@ legacy_compatibility = []
 u64_backend = ["curve25519-dalek/u64_backend"]
 u32_backend = ["curve25519-dalek/u32_backend"]
 simd_backend = ["curve25519-dalek/simd_backend"]
+# This feature exposes unverified internals for compatibility with non-spec-compliant 25519 implementations
+yolo_crypto = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,7 @@ harness = false
 # required-features = ["batch"]
 
 [features]
-default = ["std", "rand", "u64_backend"]
+default = ["std", "rand", "u64_backend", "basepoint_tables"]
 std = ["curve25519-dalek/std", "ed25519/std", "serde_crate/std", "sha2/std", "rand/std"]
 alloc = ["curve25519-dalek/alloc", "rand/alloc", "zeroize/alloc"]
 nightly = ["curve25519-dalek/nightly"]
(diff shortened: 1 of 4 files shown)


tarcieri:
See #79.

This was deliberately removed as Ed25519 is EdDSA instantiated with SHA-512 as the hash function.

As soon as you change the hash function the construction is no longer Ed25519 but some nonstandard EdDSA variant.

If you'd like a standardized EdDSA built on SHA3/SHAKE instead for whatever reason, there's always Ed448.

ryankurte:
thanks for your response!

> This was deliberately removed as Ed25519 is EdDSA instantiated with SHA-512 as the hash function.

> As soon as you change the hash function the construction is no longer Ed25519 but some nonstandard EdDSA variant.

ahh yes i missed the [prior ruling](https://github.com/dalek-cryptography/ed25519-dalek/issues/121#issuecomment-696468328), and this makes sense. i'm doing some work to replace `donna` C code (which has support for generating implementations with arbitrary hashes) for `dalek` rust bindings, so it's not a question of choosing variants so much as replicating those already in use.

from my perspective it would be really great to  have a low-friction path for the retrofitting use-case, and not to _need_ to copy / maintain / trust a second implementation of the `sign | verify | signature` code outside of this crate.

that said it's fairly straightforward to do the ol' copy-paste if y'all are unhappy with adding supporting APIs, do you have any thoughts about whether there's a path to landing something towards this (i see feature gates were discussed previously), or exposing more of the `InternalSignature` magic, to make this adventure friendlier to non-cryptographers?

rozbb:
I'm leaning towards not supporting this use case. It seems pretty niche and, if we were to do it consistently, it'd result in a lot more code noise for `sign`, `sign_prehashed`, `verify`, `verify_strict`, `verify_prehashed_strict`, and `verify_batch`.

If you want to use a small signature scheme for embedded devices, I'd look into [jq255e](https://github.com/C2SP/C2SP/blob/main/jq255.md), whose standard supports multiple hash functions, including SHA-3. It has a [Rust implementation](https://docs.rs/crrl/latest/crrl/jq255e/index.html) as well, though it only supports Blake2s at the moment.

Thoughts @tarcieri?

tarcieri:
As I stated earlier, by definition using any other hash function besides SHA-512 makes the construction a non-Ed25519 non-standard EdDSA variant, which I don't think belongs in an "ed25519-*" crate.

As a maintainer of Rust implementations of several pre-Ed25519 digital signature algorithms which allowed for hash function customizations, I think the proliferation of hash functions creates myriad incidental complexity and interoperability woes. This kind of parameterization defeats the original NaCl design goal of opinionated cryptography with a simple interface.

Supporting this also makes `digest` part of the crate's public API, where presently it is not.

rozbb:
Agreed. Closing because out of scope.

PR #203, opened 2022-06-11
Test SecretKey's Drop impl without a stack use after scope

saethlin:
Previously, this test tried to us a dangling/deallocated pointer. Fortunately there's a way to run a `Drop` impl without deallocation :)

### DIFF
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -535,16 +535,12 @@ mod test {
 
     #[test]
     fn secret_key_zeroize_on_drop() {
-        let secret_ptr: *const u8;
-
-        { // scope for the secret to ensure it's been dropped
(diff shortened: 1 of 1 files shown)


rozbb:
We removed the test entirely. @tarcieri do you know why it was there? It seems redundant to test the basic functionality provided by `zeroize`. If you think it belongs, the new code to stick at the bottom of `signing.rs` is:
```rust
```

tarcieri:
Well, it might've caught #247 😉 

I think I did port it over and convert it to use `drop_in_place` in an earlier version of #242 but there had been so many other changes I ended up redoing it instead of rebasing.

Of course, now that #247 has landed, I agree it provides limited value.

saethlin:
It's been a while since I wrote this and I'll admit I mostly don't understand this codebase. It sounds like this is no longer needed? If so can one of you close it?

PR #204, opened 2022-06-21
WIP: `alloc` free `verify_batch`

ryankurte:
hi folks ^_^

this is the beginning of an allocation free implementation of `verify_batch` via `heapless`.

i've refactored the `batch` module to support this, but am currently blocked on the requirement for `alloc` in `VartimeMultiscalarMul` implementations for `EdwardsPoint` (see `scalar_mul::strauss` and `scalar_mul::pippenger` `optional_multiscalar_mul` functions for detail), and was wondering whether y'all had thoughts on how best to approach this?

cc. @isislovecruft


ryankurte:
closing as it looks like we don't actually need this ^_^

PR #210, opened 2022-07-29
Add Keypair::from::<SecretKey>

GrishaVar:
I felt this method was missing, as `Keypair` is the most useful type for signing - it should be easy to construct.

Perhaps a `Keypair::from_secret_bytes` would be good as well, to create a `Keypair` directly from the minimal required amount of information?

### DIFF
--- a/src/keypair.rs
+++ b/src/keypair.rs
@@ -40,6 +40,14 @@ pub struct Keypair {
     pub public: PublicKey,
 }
 
+impl From<SecretKey> for Keypair {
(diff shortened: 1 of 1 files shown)


rozbb:
Resolved in #205. [Relevant lines](https://github.com/dalek-cryptography/ed25519-dalek/pull/205/files#diff-8b76c57cec65e93f54684c7f7813e49ccfc9c1cc6894184d77a09b07071b3555R45-R49)

PR #212, opened 2022-08-19
Adding verify_prehashed_strict()

orthecreedence:
Combines `verify_prehashed` and `verify_strict` to allow strict verification with prehashed values.

Wrote an integration test (`repudiation_prehash`) however it fails intermittently as the `verify_prehashed()` call sometimes fails. I will need guidance on why because I'm not familiar with the inner workings of the verification, I really just kind of merged the prehashed/strict functions together.

### DIFF
--- a/tests/ed25519.rs
+++ b/tests/ed25519.rs
@@ -186,6 +186,76 @@ mod vectors {
             pk.verify_strict(message1, &sig).is_err() || pk.verify_strict(message2, &sig).is_err()
         );
     }
+
+    fn compute_hram_prehashed(message: &[u8], context: Option<&[u8]>, pub_key: &EdwardsPoint, signature_r: &EdwardsPoint) -> Scalar {
+        let ctx: &[u8] = context.unwrap_or(b"");
+        let k_bytes = Sha512::default()
+            .chain(b"SigEd25519 no Ed25519 collisions")
+            .chain(&[1])
+            .chain(&[ctx.len() as u8])
+            .chain(ctx)
+            .chain(&signature_r.compress().as_bytes())
+            .chain(&pub_key.compress().as_bytes()[..])
+            .chain(&message);
+        let mut k_output = [0u8; 64];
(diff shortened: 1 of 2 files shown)


rozbb:
It appears the `repudiation_prehashed` test only succeeds _some_ of the time. Do you know why that is?

tarcieri:
Do these "strict" APIs really make sense for anything?

AFAIK they still have some malleability and are nonstandard. For a truly nonmalleable signature, I've seen some recent discussion on the Decaf mailing list which might be relevant.

For consensus-critical applications it seems like the ZIP-215 rules make more sense and at least have a spec to follow.

rozbb:
What malleability does `verify_strict` tolerate? I assumed from the all-blue square in the [malleability post](https://hdevalence.ca/blog/2020-10-04-its-25519am) that it affords nothing.

burdges:
You can have  public keys with 2-torsion.  I'm dubious `verify_strict` is useful, but if so then maybe they actually want a full subgroup checks on but the public key and R.  I donno..

rozbb:
Fixed the test errors. Turns out it was using the messages directly rather than their digests. I also took the opportunity to clean up the verification code so it doesn't repeat itself so many times.

@tarcieri ready for re-review

tests/ed25519.rs
```diff
@@ -116,81 +130,154 @@ mod vectors {
...
+        let mut rng = rand::rngs::OsRng;
+        let mut s_candidate = Scalar::random(&mut rng);
+        while s_candidate == Scalar::ZERO {
+            s_candidate = Scalar::random(&mut rng);
+        }
+        s_candidate
+    }
+
+    fn pick_r(s: Scalar) -> EdwardsPoint {
+        let r0 = s * ED25519_BASEPOINT_POINT;
```
tarcieri, line 192: After #251 lands we should probably look for places like this where `mul_base` could be used

tarcieri (approved):

PR #214, opened 2022-09-02
Update rand libraries

AshwinSekar:
`rand_core` is usually not compatible cross version, upgrading it here so it is usable with crates that use `rand >= 0.8`


rozbb:
Yes I apologize for the delay. We've been rushing to get a new curve25519-dalek release done. Right after that we're gonna update all the deps here and do another release.

rozbb:
Resolved in #223

PR #217, opened 2022-10-15
ping ed25519 to =1.3.0

alxiong:
as suggested by @tarcieri https://github.com/dalek-cryptography/ed25519-dalek/pull/205#issuecomment-1279609076

> The current MSRV test is failing because the ed25519 crate is Rust 2021 edition.
> The easiest way to work around that is probably to check in Cargo.lock and ensure it's computed for ed25519 v1.3.0, possibly by adding a temporary =1.3.0 version requirement when computing Cargo.lock.


rozbb:
Switching to v1.2.0 fixes the build. @tarcieri can you explain the plan for this change? Is the MSRV broken in the current release, and was this intended to fix it, or is the next release going to have the same MSRV?

tarcieri:
@rozbb I’d suggest bumping MSRV to 1.56 in the next release.

Since you just merged a breaking change (#205), maybe that should just happen now.

rozbb:
Ok agreed. So this PR is unnecessary then?

tarcieri:
If MSRV is bumped, then yeah

rozbb:
Closing in favor of discussion in https://github.com/dalek-cryptography/ed25519-dalek/pull/218

PR #218, opened 2022-10-16
Bumped MSRV to 1.56.1 and added some documentation about semver

rozbb:


tarcieri (approved):
Not sure what’s up with the benchmarks but the rest looks good

Cargo.toml
```diff
@@ -30,7 +30,7 @@ rand_core = { version = "0.5", default-features = false, optional = true }
 serde_crate = { package = "serde", version = "1.0", default-features = false, optional = true }
 serde_bytes = { version = "0.11", optional = true }
 sha2 = { version = "0.9", default-features = false }
-zeroize = { version = "~1.3", default-features = false }
+zeroize = { version = "1", default-features = false }
```
matklad, line 33: This is great, would love a release with this change!

tarcieri: See dalek-cryptography/curve25519-dalek#405

Issue #215, opened 2022-10-14
Bench issue


LumaRay:
Hi!
I get this error while performing cargo bench:

Compiling criterion v0.3.6
error[E0432]: unresolved import `ed25519_dalek::verify_batch`
  --> benches/ed25519_benchmarks.rs:24:9
   |
24 |     use ed25519_dalek::verify_batch;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `verify_batch` in the root

rozbb:
Fixed in #220

PR #220, opened 2022-10-19
Fix batch build

rozbb:
There's a few things broken with the batch build. Firstly, `batch` alone will not generate batching code, since it also requires `alloc`. This causes the following failure:

```console
$ cargo test --no-default-features --features "u64_backend batch"
error[E0425]: cannot find function `verify_batch` in this scope
   --> tests/ed25519.rs:282:22
    |
282 |         let result = verify_batch(&messages, &signatures[..], &public_keys[..]);
    |                      ^^^^^^^^^^^^ not found in this scope
```

Secondly, even when `alloc` is specified, the build will fail because `batch` relies on `rand::thread_rng`, which is feature-gated behind `std`:
```console
$ cargo build --no-default-features --features "u64_backend alloc batch"
error[E0432]: unresolved import `rand::thread_rng`
  --> src/batch.rs:34:5
   |
34 | use rand::thread_rng;
   |     ^^^^^^^^^^^^^^^^ no `thread_rng` in the root
```

Thirdly, as noted in #215, benchmarks without `batch` are broken too:
```console
$ cargo bench
error[E0432]: unresolved import `ed25519_dalek::verify_batch`
  --> benches/ed25519_benchmarks.rs:19:9
   |
19 |     use ed25519_dalek::verify_batch;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `verify_batch` in the root
```

This PR does a few things:
* General formatting (can't help this much, since my IDE does this. I'll do a proper rustfmt pass soon) and `extern crate` cleanup
* Makes it so that `batch` alone is sufficient to get batching code (i.e., `batch = ["alloc", ...]`)
* Allows benchmarks to run without `batch` enabled
* Includes regression tests for all of the above
* Makes `batch` no-std compatible by pulling in `getrandom` by default. It is also documented that `std` gives a speedup via `rand::thread_rng()`
* Updates `criterion` so we don't get the warning about the deprecated `bench_function_over_inputs` function


burdges:
Imho, there is no reason for batch verification to depend upon system randomness at all.  Instead, one should always use deterministic batching and relax the non-batch verification condition like https://github.com/ZcashFoundation/ed25519-zebra and https://github.com/penumbra-zone/ed25519-consensus See https://hdevalence.ca/blog/2020-10-04-its-25519am do, using `EdwardsPoint::is_small_order` or whatever.

It's then straightforward to implement Schnorr half-aggregation https://eprint.iacr.org/2021/350 like I did in https://github.com/w3f/schnorrkel/commit/c32761a8f0c637e6401390eb112864d2f9d62b6a 

You might want a verify function that matches some common C verify function.  If so, I'd suggest the doc comments explain that it's incompatible with batch verification, and that it not be the default called `verify`.  Also, I've never understood why `verify_strict` exists, but maybe some other name avoids confusing people.

rozbb:
I would like to have the "validation criteria" talk some time soon. Just to keep the scope of this PR narrow: I just ran some benchmarks and it appears that nondeterministic batching is at least as fast as (and sometimes 1–3% faster than) deterministic batching, over every single backend.

@tarcieri do you see any reason to keep deterministic batching? The only argument I can think of is that `merlin` will be a little bit slower on 32-bit machines due to endianness stuff. For comparison, Zebra just [takes in an RNG](https://github.com/ZcashFoundation/ed25519-zebra/blob/15e028616c6b370dee4e3399153eb8a4348fbe39/src/batch.rs#L151) and calls it a day.

burdges:
I guess non-deterministic helps if someone has botched validation criteria, which maybe unavoidable, but still merlin keeps your code simple.  I suppose this 1-3% comes from merlin being keccak for both hashing and output, while `rand::ThreadRng` is simply ChaCha12.

rozbb:
Oops, I meant to say that _deterministic_ batching is actually faster. So it'd be quite nice if we could just do that.

tarcieri:
Moving to deterministic batching seems fine by me

burdges:
Should we have the "validation criteria" talk on the IRTF mailing list?  It's afaik clear what needs to happen, but maybe some thoughts about how it should be presented to users?

rozbb:
I've opened #229 to talk about validation criteria. I think we should prioritize non-breakage over eveyrthing else. `verify_permissive()` will be a separate function.

rozbb:
@tarcieri ready for re-review

tarcieri (approved):
Something else it'd be good to fix would be generating a proper `compile_error!` if both the `batch` and `batch_deterministic` features, e.g.

```rust
#[cfg(all(feature = "batch", feature = "batch_deterministic"))]
compile_error!("the `batch` and `batch_deterministic` features are mutually exclusive");
```

Right now if you enable them both you get a confusing compile error:

```
error[E0425]: cannot find value `prng` in this scope
   --> src/batch.rs:276:31
    |
276 |         .map(|_| Scalar::from(prng.gen::<u128>()))
    |                               ^^^^ not found in this scope
```

Another way to fix this would be to completely replace `batch` with `batch_deterministic`.

rozbb:
Merging for now. Still want to look into removing one of these features

PR #223, opened 2022-11-19
Bump `curve25519-dalek` to v4.0.0-pre (via git)

tarcieri:
Also bumps these corresponding dependencies which are needed for everything to compile with this update:

- `merlin` v3.0
- `rand` v0.8
- `rand_core` v0.6
- `sha2` v0.10

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,41 +4,29 @@ on:
   push:
     branches: [ '*' ]
   pull_request:
-    branches: [ main, develop ]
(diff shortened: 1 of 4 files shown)


rozbb:
Looks good. Was there a reason this was a WIP?

tarcieri:
Just the failing CI jobs, at least one of which seems to be addressed by an open PR

tarcieri:
Removed WIP if you consider the build failures unrelated

tarcieri:
@rozbb should be ready now, even fixed the test failures

PR #224, opened 2022-11-20
PKCS#8 support

tarcieri:
~~*NOTE: depends on #222 and #223 which need to be merged first.*~~

Adds optional integration  with `ed25519::pkcs8` with support for decoding/encoding `Keypair` from/to PKCS#8-encoded documents.

~~TODO: support for decoding/encoding `PublicKey` from/to SPKI documents.~~ Added, good to go!


tarcieri:
This is feature-complete now but still needs docs

tarcieri:
This should hopefully be good to go now.

`pkcs8` has an MSRV of 1.57, so I bumped to that unilaterally for now.

Cargo.toml
```diff
@@ -12,54 +12,59 @@ keywords = ["cryptography", "ed25519", "curve25519", "signature", "ECC"]
...
-# For now, we have to bench by doing `cargo bench --features="batch"`.
-# required-features = ["batch"]
+required-features = ["batch"]
 
 [features]
-default = ["std", "rand", "u64_backend"]
-std = ["curve25519-dalek/std", "ed25519/std", "serde_crate/std", "sha2/std", "rand/std"]
-alloc = ["curve25519-dalek/alloc", "rand/alloc", "zeroize/alloc"]
+default = ["std", "rand"]
+std = ["alloc", "curve25519-dalek/std", "ed25519/std", "serde_crate/std", "sha2/std", "rand/std"]
```
tarcieri, line 54: These look like they could really benefit from namespaced features if we were okay with bumping MSRV to 1.60.

The `ed25519` crate could probably benefit from it as well. Right now it's not possible to activate `pkcs8/std` via the `ed25519` crate alone, for example.

rozbb (commented):
Looks great! Left some small comments

src/keypair.rs
```diff
@@ -443,6 +446,90 @@ impl Verifier<ed25519::Signature> for Keypair {
...
+impl TryFrom<&[u8]> for Keypair {
+    type Error = SignatureError;
+
+    fn try_from(bytes: &[u8]) -> Result<Keypair, SignatureError> {
+        Keypair::from_bytes(bytes)
+    }
+}
+
+#[cfg(feature = "pkcs8")]
+#[cfg_attr(docsrs, doc(cfg(feature = "pkcs8")))]
```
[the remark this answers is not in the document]
tarcieri: Yeah, this branch predates that, but I'll go ahead and remove them

src/keypair.rs
```diff
@@ -443,6 +446,90 @@ impl Verifier<ed25519::Signature> for Keypair {
...
+            .map_err(|_| pkcs8::Error::KeyMalformed)?;
+
+        let public = PublicKey::from(&secret);
+
+        // Validate the public key in the PKCS#8 document if present
+        if let Some(public_bytes) = pkcs8_key.public_key {
+            let pk = PublicKey::from_bytes(public_bytes.as_ref())
+                .map_err(|_| pkcs8::Error::KeyMalformed)?;
+
+            if public != pk {
```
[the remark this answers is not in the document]
tarcieri: It's not constant time, but by definition knowledge of the public key should not reveal information about the secret key

tarcieri:
Had to rebase so I guess I can go ahead and squash

tarcieri:
@rozbb this is green again

src/keypair.rs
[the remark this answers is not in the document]
rozbb: Ok tell me if this is too paranoid.  `try_from` is effectively a function of two variables: `sk` and `pk`. If an attacker has access to `pk`, then they get a side channel on the comparison `pubkey(sk) == pk`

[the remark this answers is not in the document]
tarcieri: How is that any different from saying, in a general case scenario, that an attacker has knowledge of `pk` and a function `F(sk) -> pk`?

The problem is it requires a brute force attack to do the opposite: `G(pk) -> sk`. To do that you'd need an attacker capable of efficiently solving ECDLP.

If you really want a belt-and-suspenders defense we can do a constant-time comparison, but IMO if the attacker can determine `sk` from `pk` then elliptic curve cryptography simply wouldn't work (because the attacker would be capable of solving ECDLP)

tarcieri:
Re: 5eab15d, IMO having `extern crate` anywhere than the toplevel `lib.rs` is a bit weird. I swear there used to be a warning/lint for it but perhaps I'm mistaken

tarcieri:
Ran rustfmt over it and the only problems were with the newly-added PKCS#8 code, so this is also rustfmt clean.

@rozbb ready for re-review

src/keypair.rs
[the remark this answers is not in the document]
rozbb: Oh you're right. This leaks `pubkey(sk)` via timing, but I can't imagine a scenario where a user is simultaneously hiding their pubkey and accepting half-certificates from untrusted sources. Nevermind.

src/keypair.rs
```diff
@@ -431,6 +434,83 @@ impl Verifier<ed25519::Signature> for Keypair {
     }
 }
 
+impl TryFrom<&[u8]> for Keypair {
+    type Error = SignatureError;
+
+    fn try_from(bytes: &[u8]) -> Result<Keypair, SignatureError> {
+        Keypair::from_bytes(bytes)
```
rozbb, line 441: Is there a reason to keep the `Keypair::from_bytes()` methods anymore? Documentation on methods of trait impls works how we'd want it to. Small other change is `SecretKey::from_bytes -> try_from` for consistency.

src/public.rs
```diff
@@ -350,6 +353,65 @@ impl Verifier<ed25519::Signature> for PublicKey {
     }
 }
 
+impl TryFrom<&[u8]> for PublicKey {
+    type Error = SignatureError;
+
+    fn try_from(bytes: &[u8]) -> Result<PublicKey, SignatureError> {
+        PublicKey::from_bytes(bytes)
```
rozbb, line 360: Ditto re `from_bytes`

rozbb (commented):
Left minor comment. Also since https://github.com/dalek-cryptography/curve25519-dalek/pull/470 merged, this line needs to be changed to use `Scalar::ZERO`:
https://github.com/dalek-cryptography/ed25519-dalek/blob/1a6f50a55067d3f5e212a7fcf7b61f468a169088/tests/ed25519.rs#L159

src/keypair.rs
tarcieri: Yes, per #230 it can be changed to accept a fixed-sized array or reference thereof, which can catch errors related to the input size at compile time rather than runtime. The `TryFrom<&[u8]>` impls can still accept a slice for convenience (and to keep the type signatures there simple).

As it were, a local WIP PR which I haven't pushed up yet which renames `Keypair` -> `SigningKey` and `PublicKey` -> `VerifyingKey` does just that.

rozbb: Gotcha makes sense

pinkforest:
Need to bump ed25519 to pre.1 probably ?

warning: Patch `ed25519 v2.0.0-pre.1 (https://github.com/RustCrypto/signatures.git#fd78bf79)` was not used in the crate graph.
Downloaded signature v2.0.0-pre.2

Also https://github.com/dalek-cryptography/curve25519-dalek/pull/472 needs to patch src/signature.rs now:
```
error[E0308]: mismatched types
  --> src/signature.rs:98:9
   |
97 |     match Scalar::from_canonical_bytes(bytes) {
   |           ----------------------------------- this expression has type `subtle::CtOption<Scalar>`
98 |         None => return Err(InternalError::ScalarFormatError.into()),
   |         ^^^^ expected struct `subtle::CtOption`, found enum `Option`
   |
   = note: expected struct `subtle::CtOption<Scalar>`
                found enum `Option<_>`

error[E0308]: mismatched types
  --> src/signature.rs:99:9
   |
97 |     match Scalar::from_canonical_bytes(bytes) {
   |           ----------------------------------- this expression has type `subtle::CtOption<Scalar>`
98 |         None => return Err(InternalError::ScalarFormatError.into()),
99 |         Some(x) => return Ok(x),
   |         ^^^^^^^ expected struct `subtle::CtOption`, found enum `Option`
   |
   = note: expected struct `subtle::CtOption<Scalar>`
                found enum `Option<_>`


```

pinkforest:
@rozbb - Fixed current tests failing in #239

rozbb:
I thought I was going crazy. I have pages and pages of errors when testing locally. Thankfully this shows up in the CI

pinkforest:
@rozbb ed25519 pre-release dependency needs a bump and the warning goes away as the .patch then works correctly
```
ed25519 = { version = "=2.0.0-pre.1", default-features = false }
```

Cargo.toml
```diff
@@ -49,15 +52,17 @@ required-features = ["batch"]
...
 asm = ["sha2/asm"]
 # This features turns off stricter checking for scalar malleability in signatures
 legacy_compatibility = []
+pkcs8 = ["ed25519/pkcs8"]
+pem = ["alloc", "ed25519/pem", "pkcs8"]
 
 [patch.crates-io]
 curve25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dalek.git", branch = "release/4.0" }
-ed25519 = { git = "https://github.com/RustCrypto/signatures.git"}
+ed25519 = { git = "https://github.com/RustCrypto/signatures.git" }
```
pinkforest, line 68: @rozbb ed25519 pre-release dependency needs a bump and the warning goes away as the .patch then works correctly
```
diff --git a/Cargo.toml b/Cargo.toml
index f87f430..e795560 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ features = ["nightly", "batch", "pkcs8"]
 
 [dependencies]
 curve25519-dalek = { version = "=4.0.0-pre.2", default-features = false, features = ["digest", "rand_core"] }
-ed25519 = { version = "=2.0.0-pre.0", default-features = false }
+ed25519 = { version = "=2.0.0-pre.1", default-features = false }
 merlin = { version = "3", default-features = false, optional = true }
 rand = { version = "0.8", default-features = false, optional = true }
 rand_core = { version = "0.6", default-features = false, optional = true }
```

pinkforest:
I can confirm that most of the build errors go away by bumping to .1 as I've .pathc'd 

Just leaves one batch related error after:

```
error[E0432]: unresolved import `rand::thread_rng`
  --> src/batch.rs:28:5
   |
28 | use rand::thread_rng;
   |     ^^^^^^^^^^^^^^^^ no `thread_rng` in the root

```

Which is fixed in
- #220

Cargo.toml
pinkforest: Also @rozbb the curve25519-dalek reference has to be .patched as well to `pre.3`:

```
diff --git a/Cargo.toml b/Cargo.toml
index f87f430..971328e 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 features = ["nightly", "batch", "pkcs8"]
 
 [dependencies]
-curve25519-dalek = { version = "=4.0.0-pre.2", default-features = false, features = ["digest", "rand_core"] }
-ed25519 = { version = "=2.0.0-pre.0", default-features = false }
+curve25519-dalek = { version = "=4.0.0-pre.3", default-features = false, features = ["digest", "rand_core"] }
+ed25519 = { version = "=2.0.0-pre.1", default-features = false }
 merlin = { version = "3", default-features = false, optional = true }
 rand = { version = "0.8", default-features = false, optional = true }
 rand_core = { version = "0.6", default-features = false, optional = true }
```

tarcieri:
@pinkforest `[patch.crates-io]` works just fine if Cargo.lock is checked in, which stores the upstream git SHA1 hash that was used at the time, and won't update it unless `cargo update [-p <package>]` was used explicitly.

I think the advice not to check in Cargo.lock is a bit dated now that dependabot exists, and without it builds aren't reproducible or bisectable.

pinkforest (approved):

Cargo.toml
```diff
@@ -12,18 +12,20 @@ keywords = ["cryptography", "ed25519", "curve25519", "signature", "ECC"]
...
 
 [badges]
 travis-ci = { repository = "dalek-cryptography/ed25519-dalek", branch = "master"}
 
 [package.metadata.docs.rs]
 # Disabled for now since this is borked; tracking https://github.com/rust-lang/docs.rs/issues/302
 # rustdoc-args = ["--html-in-header", ".cargo/registry/src/github.com-1ecc6299db9ec823/curve25519-dalek-0.13.2/rustdoc-include-katex-header.html"]
-features = ["nightly", "batch"]
+rustdoc-args = ["--cfg", "docsrs"]
+features = ["nightly", "batch", "pkcs8"]
```
rozbb, line 24: `nightly` doesn't exist anymore, but we can get that in another PR

PR #226, opened 2022-11-20
Fix warnings and add `-D warnings` check in CI

tarcieri:
- Eliminates dead code left over from #205
- Adds `-D warnings` in CI so that warnings are treated as errors. This ensures code must be warning-free to pass CI.


Cargo.toml
```diff
@@ -54,7 +54,7 @@ std = ["curve25519-dalek/std", "ed25519/std", "serde_crate/std", "sha2/std", "ra
 alloc = ["curve25519-dalek/alloc", "rand/alloc", "zeroize/alloc"]
 nightly = ["curve25519-dalek/nightly"]
 serde = ["serde_crate", "serde_bytes", "ed25519/serde"]
-batch = ["merlin", "rand"]
+batch = ["merlin", "rand/std"]
```
tarcieri, line 57: It needs `std` for `rand::thread_rng`

benches/ed25519_benchmarks.rs
```diff
@@ -57,6 +57,8 @@ mod ed25519_benches {
     fn verify_batch_signatures(c: &mut Criterion) {
         static BATCH_SIZES: [usize; 8] = [4, 8, 16, 32, 64, 96, 128, 256];
 
+        // TODO: use BenchmarkGroups instead.
+        #[allow(deprecated)]
```
rozbb, line 61: I address this and the `std` stuff in #220. Will rebase soon

rozbb:
Thanks! The warnings have been a bit annoying

PR #227, opened 2022-11-21
Tune up CI configuration

tarcieri:
- Consolidate `test` jobs: this allows reusing intermediate artifacts between tests which should improve build times, and also make it easier to test additional features in the future
- Switch to `dtolnay/rust-toolchain` for setting up toolchain
- Bump checkout to `actions/checkout@3`
- Switch to `run` directives for invoking Cargo: it's more straightforward to just call Cargo than use a DSL from an unmaintained action, and eliminates the 3rd party dependency

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,7 +19,6 @@ jobs:
           # 32-bit target
           - target: i686-unknown-linux-gnu
             deps: sudo apt update && sudo apt install gcc-multilib
-
           # 64-bit target
           - target: x86_64-unknown-linux-gnu
     steps:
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -44,9 +44,7 @@ toml = { version = "0.5" }
 [[bench]]
 name = "ed25519_benchmarks"
 harness = false
-# This doesn't seem to work with criterion, cf. https://github.com/bheisler/criterion.rs/issues/344
-# For now, we have to bench by doing `cargo bench --features="batch"`.
-# required-features = ["batch"]
+required-features = ["batch"]
```
tarcieri, line 47: Per the (formerly) linked issue, this is working as expected

.github/workflows/rust.yml
```diff
@@ -19,102 +19,39 @@ jobs:
...
-      with:
-        profile: minimal
-        toolchain: 1.56.1
-        override: true
-    - uses: actions-rs/cargo@v1
-      with:
-        command: build
+    - uses: actions/checkout@v3
+    - uses: dtolnay/rust-toolchain@1.56.1
+    - run: cargo build --benches
```
rozbb, line 49: Why does the MSRV have to build the benches?

.github/workflows/rust.yml
```diff
@@ -19,102 +19,39 @@ jobs:
...
-        override: true
-    - uses: actions-rs/cargo@v1
-      with:
-        command: test
-        args: --features "batch_deterministic"
+      - uses: actions/checkout@v2
+      - uses: dtolnay/rust-toolchain@nightly
+      - run: cargo test --features simd_backend
 
   msrv:
```
rozbb, line 43: Could you make this equivalent to the [more thorough](https://github.com/dalek-cryptography/curve25519-dalek/blob/release/4.0/.github/workflows/rust.yml#L72) MSRV build test from curve25519-dalek?

rozbb (changes requested):

.github/workflows/rust.yml
tarcieri: Fixed in ac86d13

tarcieri:
@rozbb should be good now

Issue #225, opened 2022-11-20
Use `SigningKey`/`VerifyingKey` naming convention?


tarcieri:
In the @RustCrypto project we have adopted a convention of naming signature keys `SigningKey` and `VerifyingKey` which describes keys by their *roles* in a signature system.

The basic pitch is that such names are clearer and more intuitive to users who don't necessarily have a background in cryptography.

Would it make sense to adopt similar names in this crate?

rozbb:
This sounds like a reasonable thing to do for 2.0

burdges:
You're not bothered and/or like the overlap with trusted setups?

burdges:
Yeah.  I miss-phrased the question.  I meant: Did you think about it?  I've no opinion really.

tarcieri:
They serve similar roles, but typical signature algorithms are proving knowledge of the signature key

tarcieri:
A semi-related question (which perhaps deserves its own issue) is if the `SecretKey` and `Keypair` types should be combined into a single type, e.g. `SigningKey`.

Especially after #205, there is little value in keeping them separate, as the related API surface was exposed.

I would suggest adopting the RFC8032 definition of a secret key being 32-bytes. IMO the 64-byte "keypair" types (which contain the 32-byte secret key + 32-byte public key which can be derived from the secret key) is something of an accidental relic of the `ref10` implementation of Ed25519 from SUPERCOP being somewhat haphazardly repurposed into production code, i.e. the C struct representing a keypair being effectively transmuted into bytes and considered a key.

The recent public key oracle attack shows that not only is using that 64-bit serialization as a wire format needlessly wasteful, since the public key can be derived from the secret key, but it also introduces potential misuse. If we were to keep it around (i.e. the current `Keypair::from_bytes`), e.g. as `SigningKey::from_keypair_bytes` or so forth, I would suggest deprecating it and trying to move people away from that format towards RFC8032 secret keys.

rozbb:
I'm pro removing `Keypair`. Whenever possible, I'd like to avoid defining semantics for data structures whose functions overlap. How will the deprecation interact with #224?

tarcieri:
It'd probably be good to land #224 first before making these changes.

I think really it'd be folding the existing `Keypair` type and `SecretKey` type into a single `SigningKey` type. The signing functionality is impl'd on `Keypair` because it needs access to the public key to deterministically compute the unique per-message pseudorandom scalar used to compute the signature.

[...]

PR #230, opened 2022-12-01
Change from_bytes methods to take fixed-size array argument

mina86:
Change from_bytes methods of all the types to take `&[u8; N]` argument
(with `N` appropriate for given type) rather than `&[u8]`.  This does
a few things:
- it makes SecretKey::from_bytes infallible which will simplify call
  sites (if the call site already has properly-sized array),
- harmonises the convention with ed25519::Signature::from_bytes method
  and
- encodes the size of the bytes representation in type system which
  means that users can assert it at compile time.

To still allow creating the objects from a slice whose size is checked
at run time, introduce `TryFrom<&[u8]>` implementations for affected
public types.

This is an API breaking change.  The simplest way to update existing
code is to replace Foo::from_bytes with Foo::try_from.  This should
cover majority of uses.

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -162,30 +162,12 @@ impl InternalSignature {
     /// only checking the most significant three bits.  (See also the
     /// documentation for `PublicKey.verify_strict`.)
     #[inline]
-    pub fn from_bytes(bytes: &[u8]) -> Result<InternalSignature, SignatureError> {
-        if bytes.len() != SIGNATURE_LENGTH {
-            return Err(InternalError::BytesLengthError {
-                name: "Signature",
-                length: SIGNATURE_LENGTH,
-            }
-            .into());
-        }
-        let mut lower: [u8; 32] = [0u8; 32];
-        let mut upper: [u8; 32] = [0u8; 32];
-
-        lower.copy_from_slice(&bytes[..32]);
-        upper.copy_from_slice(&bytes[32..]);
-
-        let s: Scalar;
-
-        match check_scalar(upper) {
-            Ok(x) => s = x,
-            Err(x) => return Err(x),
-        }
-
+    pub fn from_bytes(bytes: &[u8; SIGNATURE_LENGTH]) -> Result<InternalSignature, SignatureError> {
+        // TODO: Use bytes.split_array_ref once it’s in MSRV.
+        let (lower, upper) = bytes.split_at(32);
         Ok(InternalSignature {
-            R: CompressedEdwardsY(lower),
-            s: s,
+            R: CompressedEdwardsY(lower.try_into().unwrap()),
+            s: check_scalar(upper.try_into().unwrap())?,
         })
     }
 }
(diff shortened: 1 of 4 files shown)


src/keypair.rs
```diff
@@ -49,6 +49,24 @@ impl From<SecretKey> for Keypair {
...
+        let secret = SecretKey::from(secret);
+        let public = PublicKey::try_from(public)?;
+
+        if public != (&secret).into() {
+            Err(InternalError::MismatchedKeypairError.into())
+        } else {
+            Ok(Keypair { secret, public })
+        }
+    }
+}
```
tarcieri, line 68: This is copy/paste of `Keypair::from_bytes`:

https://github.com/dalek-cryptography/ed25519-dalek/blob/9638ab4/src/keypair.rs#L94-L110

At the very least one should be composed in terms of the other.

But also as I [discussed in #225](https://github.com/dalek-cryptography/ed25519-dalek/issues/225#issuecomment-1322358913), we should probably get rid of the 64-byte keypair APIs, or at least deprecate them, as they don't follow the RFC8032 definition of a secret key and include redundant information which provides no additional value and only permits potential misuse.

mina86: > At the very least one should be composed in terms of the other.

They are.  This commit changes from_bytes to check length and then call try_from.

I actually don’t care about Keypair, but I figured that if I’m adding the conversion to SecretKey and PublicKey it makes sense for completeness to have it in Keypair as well.

tarcieri:
In general I'd suggest holding off on this until #225 is addressed, since that proposes to change the public APIs for keys

tarcieri:
This was also an interesting read on the actual costs of length checks: https://blog.readyset.io/bounds-checks/

tl;dr: not much, especially in cases like this

mina86:
The main thing is that creation of SecretKey is now infallible.  Another is that the expected length is now expressed in type system.  Performance isn’t really my main concern.

tarcieri:
I'd probably suggest making the `TryFrom` impls take `&[u8]` and using inherent methods (i.e. `from_bytes`) for the fixed-sized inputs. That's the approach we use with @RustCrypto.

mina86:
I can change `from_bytes` if you prefer though that is an API breaking change.

mina86:
So I’m looking at [ed25519::Signature::from_bytes from a preview ed25519 crate](https://docs.rs/ed25519/2.0.0-pre.0/ed25519/struct.Signature.html#method.from_bytes) and it has `from_bytes(&[u8])` so should this follow that instead? Or should ed25519 crate be changed as well?

tarcieri:
I'd be open to changing that method

mina86:
Updated as per suggestion.  Happy to wait till #225 gets resolved.

tarcieri:
#225 is resolved

tarcieri:
@mina86 you need to change the base to `release/2.0`

tarcieri (approved):

PR #235, opened 2022-12-08
Use namespaced/weak features; MSRV 1.60

### LINKED ISSUE #234: Reduce proliferation of nondefault features
Some features are accidentally activated transitively. For example, `alloc -> rand/alloc`, which activates the `rand` feature. This is a problem if we want to provide semver guarantees for default-on features.

Relatedly, `cargo test --no-default-features` fails on `release/2.0` currently because a lot of our tests use `generate()` which requires `rand` (or, consequently, `alloc, `std`,`batch`, or `batch_deterministic`). I think a `mock.rs` module would be useful here, but we have a lot of doctests that would be affected and it'd look nasty. Maybe worth it regardless though.

tarcieri:
This enables activating the `alloc` and `std` features without unnecessarily pulling in optional dependencies like `rand` and `serde`.

It also fixes tests for `--no-default-features` (w\ `--lib` only)

Closes #234 (I think?)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,7 +26,8 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: rustup target add ${{ matrix.target }}
       - run: ${{ matrix.deps }}
-      - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc --lib
       - run: cargo test --target ${{ matrix.target }}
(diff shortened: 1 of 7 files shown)


rozbb:
LGTM. One thing I wanna do is go through all the `std` flags and see which ones we actually use. My guess is most do nothing

tarcieri:
@rozbb the main use is enabling the `std::error::Error` impls for the respective error types for these dependencies, however there is [ongoing work to move that to `core`](https://github.com/rust-lang/project-error-handling/issues/11).

However, there is at least one legitimate use for `pkcs8`: it enables support for reading/writing keys from/to files, including setting the file permissions for private keys appropriately:

https://docs.rs/pkcs8/latest/pkcs8/trait.EncodePrivateKey.html#method.write_pkcs8_der_file

PR #236, opened 2022-12-10
Cargo.toml: compatibility updates for `curve25519-dalek` and `ed25519`

tarcieri:
This PR contains the necessary updates for `ed25519-dalek` to be compatible with upstream `curve25519-dalek` and `ed25519` as sourced from git.

# curve25519-dalek:

- Enables `digest` and `rand_core` features
- Removes transitive `nightly`, `simd_backend`, and `std` features

# ed25519:

- `AsRef` impl for `Signature` has been removed; uses `to_bytes`
- Uses `try_from` for `InternalSignature` conversion

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,11 +39,11 @@ jobs:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@nightly
       - env:
-          RUSTFLAGS: "-C target_feature=+avx2"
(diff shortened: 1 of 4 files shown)


rozbb:
LGTM for now. We'll have to turn `digest` and `rand` into optional deps with weak features later on

tarcieri:
`digest` is encapsulated so it isn’t part of the public API

But the point stands for `rand_core`

PR #239, opened 2022-12-13
Propagate CtOption and ZERO from curve/4.0

pinkforest:
- Assuming check_scalar does not have to be CT I used dirty .into() ?


tarcieri (approved):

tarcieri:
Note: checking in Cargo.lock would prevent these kinds of breakages

rozbb:
Sorry for the bother. These changes are now covered in #224, but the build is still broken 😕

Issue #228, opened 2022-11-21
Merlin dep is broken on big-endian machines


rozbb:
`merlin` fails tests on big endian machines https://github.com/dalek-cryptography/merlin/pull/60#issuecomment-697043828, due to some underlying casting that isn't respectful of endianness. It _probably_ isn't a security issue, but I don't like it.

Currently there are two PRs on two different `merlin` repos that fix this.
* In [`dalek-cryptography/merlin`](https://github.com/dalek-cryptography/merlin/pull/65) this replaces the STROBE impl entirely with `strobe-rs`
* In [`zkcrypto/merlin`](https://github.com/zkcrypto/merlin/pull/5) this fixes the incorrect cast that the keccakf function does

If we don't have write access to either repo, we might have to fork and maintain yet another :(

burdges:
I'd wager deterministic batch verification is not used by many people yet, and ed25519-zebra uses Sha512 for this.

rozbb:
> this crate gets it wrong anyways
How so? It would be really bad if we published something broken

> I'd swap in Sha512 now
In place of what? Merlin? Using Merlin is far more secure than making my own transcript mechanism. STROBE has figured all of these things out for me.

burdges:
You need the validity criteria to be permissive like ed25519-zebra, not merely in the batch verification but in the stand along verification too.  It's not I think deterministic batch verification otherwise.

It's merely an xof that sha512 lacks, not the full transcript functionality, so you'd hash the batch and then expand via cloning the sha512 state or chacha or whatever ed25519-zebra does.  I like merlin but it's a dependency and ed25519 uses sha512

tarcieri:
ed25519-dalek has moved to: https://github.com/dalek-cryptography/curve25519-dalek/tree/main/ed25519-dalek

If you feel this issue has not yet been addressed, please request this issue be moved to that repo by leaving a comment, or open a new issue on that repo.

PR #240, opened 2022-12-13
Error with `batch` and `batch_deterministic` feature combo

pinkforest:
This PR depends on
- #239 

Currently the error if both are activated is a random one with cfg()'s that didn't expect someone would use both.

Also simplified cfg() in batch.rs

This creates proper error if both are activated.

foobar@foobar-MS-7C77:~/cc/ed25519-dalek$ cargo test --all-features
```
   Compiling ed25519-dalek v1.0.1 (/home/foobar/cc/ed25519-dalek)
error: Features batch and batch_deterministic are mutually exclusive.
   --> src/lib.rs:236:1
    |
236 | compile_error!("Features batch and batch_deterministic are mutually exclusive.");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -24,10 +24,10 @@ pub use curve25519_dalek::digest::Digest;
 
 use merlin::Transcript;
 
-#[cfg(all(feature = "batch", not(feature = "batch_deterministic")))]
+#[cfg(not(feature = "batch_deterministic"))]
 use rand::thread_rng;
 use rand::Rng;
-#[cfg(all(not(feature = "batch"), feature = "batch_deterministic"))]
+#[cfg(feature = "batch_deterministic")]
 use rand_core;
 
 use sha2::Sha512;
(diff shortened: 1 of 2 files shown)


tarcieri:
Is there any reason to keep `batch` instead of replacing it entirely with `batch_deterministic`?

pinkforest:
to keep `rand_core` optional I guess ?

pinkforest:
@tarcieri https://github.com/dalek-cryptography/ed25519-dalek/issues/228#issuecomment-1321759853

Apparently the deterministic one is unstable ?

pinkforest:
Oh there is also this thread yes:
- https://github.com/dalek-cryptography/ed25519-dalek/pull/220#issuecomment-1283703908

So shall I get rid of old batch here ?

tarcieri:
I think that’s definitely worth discussing

pinkforest:
Closed in favor of #220 (all sorts of batch fixes) and #235 (compile error)

PR #242, opened 2022-12-13
Rename `Keypair` => `SigningKey`; `PublicKey` => `VerifyingKey`

tarcieri:
~~*NOTE: includes #224 which should be merged first*~~ Merged!

As proposed in #225, renames key types after their roles:

- `SigningKey` produces signatures
- `VerifyingKey` verifies signatures

The `SecretKey` type is changed to a type alias for `[u8; 32]`, which matches the RFC8032 definition:

https://www.rfc-editor.org/rfc/rfc8032#section-5.1.5

> The private key is 32 octets (256 bits, corresponding to b) of
> cryptographically secure random data.


tarcieri:
Note: git is pretty dumb about commits that rename files and make changes to the renamed files.

To properly preserve the history it'd be better to have two commits: one which renames:

- `secret.rs` => `signing.rs`
- `public.rs` => `verifying.rs`

tarcieri:
I've deliberately split this into two commits.

17cc7bba09c0f287b813426c42abbb19d09aeb0a renames the modules without making other changes. This should be sufficient to preserve the git history (i.e. for `git blame`)

fd49c82dc8227971af3156c4283680f2d5c78cee (at least for now) contains the code changes.

They should get merged as two separate commits to ensure the history of the files isn't lost.

pinkforest (approved):
lgtm - I've also benchmarked with this PR for 2.x here - there is also 1.x result:
- Rendered: https://rustcrypto.github.io/benchmarks/ed25519/dalek/
- Source: https://github.com/RustCrypto/benchmarks/tree/main/ed25519/dalek/

**On nightly-x86_64-unknown-linux-gnu @ 10700K**

Criterion run logs
- [1.x](https://github.com/RustCrypto/benchmarks/tree/main/ed25519/dalek/1.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/)
- [2.x](https://github.com/RustCrypto/benchmarks/tree/main/ed25519/dalek/2.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/)

Criterion reports
- [u32](https://rustcrypto.github.io/benchmarks/ed25519/dalek/2.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/criterion/u32-baseline/report/index.html) - baseline
- [u64](https://rustcrypto.github.io/benchmarks/ed25519/dalek/2.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/criterion/u64-delta-u32/report/index.html) - delta over u32
- [fiat_u32](https://rustcrypto.github.io/benchmarks/ed25519/dalek/2.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/criterion/fiat_u32-delta-u64/report/index.html) - delta over u64
- [fiat_u64](https://rustcrypto.github.io/benchmarks/ed25519/dalek/2.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/criterion/fiat_u64-delta-fiat_u32/report/index.html) - delta over fiat_u32
- [avx2](https://rustcrypto.github.io/benchmarks/ed25519/dalek/2.x/nightly-x86_64-unknown-linux-gnu/2022-14-1858-000/criterion/avx2-delta-fiat_u64/report/index.html) - delta over fiat_u64

src/lib.rs
```diff
@@ -114,24 +114,22 @@
...
 //! # use ed25519_dalek::{PUBLIC_KEY_LENGTH, SECRET_KEY_LENGTH, KEYPAIR_LENGTH, SIGNATURE_LENGTH};
-//! # fn do_test() -> Result<(SecretKey, PublicKey, Keypair, Signature), SignatureError> {
+//! # fn do_test() -> Result<(SigningKey, VerifyingKey, Signature), SignatureError> {
 //! # let mut csprng = OsRng{};
-//! # let keypair_orig: Keypair = Keypair::generate(&mut csprng);
+//! # let signing_key_orig: SigningKey = SigningKey::generate(&mut csprng);
 //! # let message: &[u8] = b"This is a test of the tsunami alert system.";
-//! # let signature_orig: Signature = keypair_orig.sign(message);
-//! # let public_key_bytes: [u8; PUBLIC_KEY_LENGTH] = keypair_orig.public_key().to_bytes();
-//! # let secret_key_bytes: [u8; SECRET_KEY_LENGTH] = keypair_orig.secret_key().to_bytes();
```
rozbb, line 125: So there used to be a way to get a `SecretKey` from a `Keypair` and now there isn't? This seems fine. Is there a particular reason that was there in the first place?

tests/ed25519.rs
```diff
@@ -85,20 +84,21 @@ mod vectors {
...
         let msg_bytes: Vec<u8> = FromHex::from_hex(message).unwrap();
         let sig_bytes: Vec<u8> = FromHex::from_hex(signature).unwrap();
 
-        let secret: SecretKey = SecretKey::from_bytes(&sec_bytes[..SECRET_KEY_LENGTH]).unwrap();
-        let expected_public: PublicKey =
-            PublicKey::from_bytes(&pub_bytes[..PUBLIC_KEY_LENGTH]).unwrap();
-        let keypair: Keypair = Keypair::from(secret);
-        assert_eq!(expected_public, keypair.public_key());
+        let signing_key: SigningKey =
+            SigningKey::try_from(&sec_bytes[..SECRET_KEY_LENGTH]).unwrap();
```
rozbb, line 98: Why does this work? Aren't `SecretKey` and `SigningKey` different?

src/signing.rs
```diff
@@ -512,23 +592,227 @@ impl TryFrom<pkcs8::PrivateKeyInfo<'_>> for Keypair {
...
 #[cfg(feature = "serde")]
-impl<'d> Deserialize<'d> for Keypair {
+impl<'d> Deserialize<'d> for SigningKey {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
         D: Deserializer<'d>,
     {
         let bytes = <SerdeByteBuf>::deserialize(deserializer)?;
-        Keypair::from_bytes(bytes.as_ref()).map_err(SerdeError::custom)
+        Self::try_from(bytes.as_ref()).map_err(SerdeError::custom)
```
rozbb, line 611: Just for recordkeeping, this is the diff I got from the previous version of this block of code in `signing.rs`:
```diff
0a1
>
23c24
< // bitcoind, a user might wish to have one master keypair from which others are
---
> // bitcoind, a user might wish to have one master signing key from which others are
51c52
< impl<'a> From<&'a SecretKey> for ExpandedSecretKey {
---
> impl From<&SecretKey> for ExpandedSecretKey {
67c68
<     fn from(secret_key: &'a SecretKey) -> ExpandedSecretKey {
---
>     fn from(secret_key: &SecretKey) -> ExpandedSecretKey {
73c74
<         h.update(secret_key.as_bytes());
---
>         h.update(secret_key);
93c94
<     pub(crate) fn sign(&self, message: &[u8], public_key: &PublicKey) -> ed25519::Signature {
---
>     pub(crate) fn sign(&self, message: &[u8], verifying_key: &VerifyingKey) -> ed25519::Signature {
104c105
<         R = (&r * &constants::ED25519_BASEPOINT_TABLE).compress();
---
>         R = (&r * &ED25519_BASEPOINT_TABLE).compress();
108c109
<         h.update(public_key.as_bytes());
---
>         h.update(verifying_key.as_bytes());
125c126
<     /// * `public_key` is a [`PublicKey`] which corresponds to this secret key.
---
>     /// * `verifying_key` is a [`VerifyingKey`] which corresponds to this secret key.
141c142
<         public_key: &PublicKey,
---
>         verifying_key: &VerifyingKey,
188c189
<         R = (&r * &constants::ED25519_BASEPOINT_TABLE).compress();
---
>         R = (&r * &ED25519_BASEPOINT_TABLE).compress();
196c197
<             .chain_update(public_key.as_bytes())
---
>             .chain_update(verifying_key.as_bytes())
203d203
<     }
```

rozbb (commented):
LGTM, had like 1 clarifying question

tests/ed25519.rs
tarcieri: They're different: `SecretKey` is just a `[u8; 32]` alias. That said, I'm not sure why it's relevant here?

I switched `SigningKey` to use the 32-byte RFC8032 serialization of what it calls a "private key".

The 64-byte serialization which includes the public key as well is preserved as `from_keypair_bytes` and `to_keypair_bytes`, however I think we should discourage those as they are a nonstandard serialization based on effectively transmuting the ref10 key type, because SUPERCOP didn't need or care about serialization APIs. They only make the key longer and leave a sharp edge for the public key oracle.

src/signing.rs
tarcieri: Aah sorry, some of the comments might have some find/replace changes which don't make sense.

I tried to look for as many of those as I could find but it seems like there are a few I missed.

tarcieri: Reverted "master signing key" => "master keypair" in d6b0876

tarcieri:
@rozbb any other issues?

pinkforest:
:partying_face: :tada:

PR #251, opened 2022-12-28
Add on-by-default `fast` crate feature for gating basepoint tables

tarcieri:
~~*NOTE: depends on https://github.com/dalek-cryptography/curve25519-dalek/pull/489 which needs to be merged first*~~ Merged!

~~*NOTE: now depends on #260 which needs to be merged first*~~ Merged!

Disabling the feature reduces overall code size at the cost of performance, which is useful for e.g. embedded users.

This feature transitively enables the `basepoint-tables` feature in `curve25519-dalek`.

~~TODO: fallback support for using variable-base scalar multiplication by the basepoint when the `basepoint-tables` feature is disabled.~~ Added in 8f56c5c

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -35,9 +35,10 @@ jobs:
       - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc,fast,rand_core --lib --tests
       - run: cargo test --target ${{ matrix.target }}
       - run: cargo test --target ${{ matrix.target }} --features batch
-      - run: cargo test --target ${{ matrix.target }} --features "digest rand_core"
+      - run: cargo test --target ${{ matrix.target }} --features digest,rand_core
       - run: cargo test --target ${{ matrix.target }} --features serde
       - run: cargo test --target ${{ matrix.target }} --features pem
+      - run: cargo test --target ${{ matrix.target }} --all-features
 
   build-simd:
     name: Test simd backend (nightly)
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -26,8 +26,13 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: rustup target add ${{ matrix.target }}
       - run: ${{ matrix.deps }}
-      - run: cargo test --target ${{ matrix.target }} --no-default-features --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --lib --tests
       - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features fast --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features rand_core --lib --tests
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc,rand_core --lib --tests
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features fast,rand_core --lib --tests
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc,fast,rand_core --lib --tests
       - run: cargo test --target ${{ matrix.target }}
       - run: cargo test --target ${{ matrix.target }} --features batch
       - run: cargo test --target ${{ matrix.target }} --features serde
--- a/src/verifying.rs
+++ b/src/verifying.rs
@@ -144,7 +144,8 @@ impl VerifyingKey {
         VerifyingKey(compressed, point)
     }
 
-    // A helper function that computes H(R || A || M) as well as its prehashed version
+    // A helper function that computes H(R || A || M). If `context.is_some()`, this does the
+    // prehashed variant of the computation using its contents.
     #[allow(non_snake_case)]
     fn compute_challenge(
         context: Option<&[u8]>,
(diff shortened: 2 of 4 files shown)


.github/workflows/rust.yml
```diff
@@ -26,8 +26,11 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: rustup target add ${{ matrix.target }}
       - run: ${{ matrix.deps }}
-      - run: cargo test --target ${{ matrix.target }} --no-default-features --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --lib --tests
       - run: cargo test --target ${{ matrix.target }} --no-default-features --features alloc --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features basepoint-tables --lib
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features rand --lib --tests
+      - run: cargo test --target ${{ matrix.target }} --no-default-features --features basepoint-tables,rand --lib --tests
```
tarcieri, line 33: Running much of the test suite depends on `rand`

tarcieri:
It might be worth giving this feature a name which makes its function more readily apparent to non-expert end-users, perhaps something like `fast`.

src/mul_base.rs
```diff
@@ -0,0 +1,25 @@
...
+    #[cfg(not(feature = "fast"))]
+    {
+        scalar * constants::ED25519_BASEPOINT_POINT
+    }
+
+    #[cfg(feature = "fast")]
+    {
+        scalar * constants::ED25519_BASEPOINT_TABLE
+    }
+}
```
tarcieri, line 25: Thinking about this a bit more, it seems like this is a concern which could be pushed down into `curve25519-dalek`

tarcieri: Moved to https://github.com/dalek-cryptography/curve25519-dalek/pull/489

tarcieri:
Not sure why the build is failing. I split out #260 to handle updating to the latest upstream `curve25519-dalek` changes as an intermediate debugging step.

tarcieri:
@rozbb this is ready now

rozbb:
Just ran some benchmarks. It looks like we get massive (64%) speedups in keygen and signing. But we get a 1-4% slowdown in batch verification, and no effect at all for single verification. I'll run it on another machine as well. But if this trend holds up, do you wanna rename to `fast-sign`?

tarcieri:
Haha what? That's surprising.

If that's really the case then batch verification should always use `... * ED25519_BASEPOINT_POINT` if using `mul_base` has a negative impact on performance when the basepoint tables feature is enabled

rozbb:
Yeah makes sense. First I'll test how the feature interacts with the simd backends.

rozbb:
Yup same slowdown even with AVX2 SIMD. And actually, single signature verification is 5-9% slower with `fast` enabled. So there's no situation wherein verification benefits from the tables. This is _really_ surprising.

 I'll do the following: rename the feature to `fast-sign`, explain it in the README (maybe after merging the other README PR), and make all verification use table-less scalar mult.

tarcieri:
I still think we should remove `mul_base` anywhere it isn’t helping and keep the feature name as `fast`.

If there are slowdowns occurring in places where `mul_base` isn’t used, *that* is a real mystery

tarcieri:
Perhaps we should open an issue on `curve25519-dalek`?y

Tracing the code I'm perplexed how this feature would have any effect on verification, either single signature or batch verification, as as far as I can tell neither use the basepoint tables? (although there are many backends and maybe I haven't grepped/traced properly)

Batch verification calls `EdwardsPoint::optional_multiscalar_mul` which starts with the basepoint constant (i.e. point, not table).

Single signature verification calls `EdwardsPoint::vartime_double_scalar_mul_basepoint` which calls `scalar_mul::vartime_double_base::mul` which uses a completely different basepoint table `AFFINE_ODD_MULTIPLES_OF_BASEPOINT` located in `backend::serial::u64::constants` (which, sidebar, should probably also be gated by the precomputed tables feature, whatever we end up calling it)

rozbb:
Yeah I ran into the same thing when tracing this down. I have no idea what's causing this. I'll spend a little more time tracing things

src/verifying.rs
```diff
@@ -166,6 +167,22 @@ impl VerifyingKey {
...
+    // Helper function for verification. Computes the _expected_ R component of the signature. The
+    // caller compares this to the real R component.  If `context.is_some()`, this does the
+    // prehashed variant of the computation using its contents.
+    #[allow(non_snake_case)]
+    fn recompute_r(
+        &self,
+        context: Option<&[u8]>,
+        signature: &InternalSignature,
+        M: &[u8],
+    ) -> EdwardsPoint {
```
rozbb, line 179: Note: this will likely return a `CompressedEdwardsY` in the process of fixing #267 . Documenting the comment here:
```
Note that this returns the compressed form of R and the caller does a byte comparison.
This means that all our verification functions do not accept non-canonically encoded R values.
See the validation criteria blog post for more details:
    https://hdevalence.ca/blog/2020-10-04-its-25519am
```

tarcieri:
@rozbb let me update `curve25519-dalek` in this PR now that the feature name has changed to `precomputed-tables`

tarcieri:
Updated and it's green again

Issue #20, opened 2018-04-29
Ed25519 malleability vs libsodium


aep:
Hi,

i'm still confused if malleability is a problem or not. 

"A signature scheme is malleable if, given a signature σ on a message m, one can
efficiently derive a signature σ' on a message m' = T(m) for an allowed transformation T."

The m' = T(m)  is where i get lost. If there isn't any additional MAC, doesnt that mean any T is valid and therefor the signature schema is broken?

Even if not, libsodium has added this change

https://github.com/jedisct1/libsodium/issues/125

And as far as i my limited understanding goes this may lead to ed25519-dalek producing signatures rejected by libsodium. Would it be worth following libsodium here, since it's very popular and incompatibilities may be rather unexpected?

isislovecruft:
Hi @aep!

Thanks for asking about this! So the malleability in question is due to the fact that curve25519 does not have prime order (meaning that the number of points on the curve is not prime). Instead, the order is `l * c` where `l` (henceforth written as `\ell` in order to distinguish it from the number one) is the order of the base point, i.e. `\ell = 2^252 + 27742317777372353535851937790883648493` (which is prime) and the cofactor `c = 8`. The cofactor is the order of a torsion component to each curve point. For visual people, if you imagine a prime-order group as a (very) large cog, then this torsion component would be like a smaller cog, something like this (except that the big cog should have `\ell` teeth and little cog should only have `c` teeth):

![small-cog-large-cog](https://user-images.githubusercontent.com/907241/40150059-fae37124-5966-11e8-8bdd-077725e57ce0.jpg)

And then imagine that to produce a point on the curve, it depends on how each cog is (independently) rotated. **This means that every time any cryptographic primitive depends upon something being within a prime-order group, there are potential eight "equal" things in curve25519**.

This comes into play w.r.t. signatures in two ways:

1. Public ed25519 keys are points on the extended twisted Edwards form of the curve.
2. ed25519 signatures are of the form `(R, s)`, where `R` is again a point in extended twisted Edwards format and `s` is a scalar (for this curve, an integer modulo `\ell = 2^252 + 27742317777372353535851937790883648493`).

[...]

PR #253, opened 2023-01-02
Validation criteria tests

### LINKED ISSUE #229: Write unit tests for validation criteria, and write `verify_permissive()`
Currently our validation criteria are implicit and unique among all the ed25519 implementations. We should:

* Measure how `verify()` performs on the [CCTV test vectors](https://github.com/C2SP/CCTV/blob/main/ed25519vectors/ed25519vectors.json) and turn that into asserts so we can catch validation-breaking changes in CI
* Measure how `verify_strict()` performs on the same test vectors. It should reject every one of them.
* Write and expose a `verify_permissive()` function, and ensure it permits all the standard-abiding test vectors. The reason for a `verify_permissive()` is mostly due to Henry's reasoning in the [validation criteria post](https://hdevalence.ca/blog/2020-10-04-its-25519am): it is really weird to have individual signature verification to differ from batch verification. It's also useful for consensus purposes, which is why `ed25519-zebra` exists at all.

rozbb:
This code determines and tests the validation edge cases that our `verify()` and `verify_strict()` functions permit. The test vectors come from [CCTV](https://github.com/C2SP/CCTV/tree/5ea85644bd035c555900a2f707f7e4c31ea65ced/ed25519vectors).

The README there has a more detailed explanation of what exactly is being tested, but the gist is: there are some weird public keys and signatures which have more than one representation. This is classified into a number of edge cases, which these test vectors check.

One interesting finding here: Fillipo claims in the README that `verify_strict()` rejects all the test vectors with edge cases listed, but that's not true. `verify_strict()` passes test case 7, which is `low_order_component_R & low_order_component_A` (i.e., has a torsion point added to `R` and `A`). Is this a regression? Was the claim ever true?

Partially resolves #229

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ documentation = "https://docs.rs/ed25519-dalek"
 keywords = ["cryptography", "ed25519", "curve25519", "signature", "ECC"]
 categories = ["cryptography", "no-std"]
 description = "Fast and efficient ed25519 EdDSA key generations, signing, and verification in pure Rust."
-exclude = [ ".gitignore", "TESTVECTORS", "res/*" ]
+exclude = [ ".gitignore", "TESTVECTORS", "VALIDATIONVECTORS", "res/*" ]
 rust-version = "1.60"
 
 [badges]
--- a/VALIDATIONVECTORS
+++ b/VALIDATIONVECTORS
@@ -1,0 +1,11136 @@
+[
+	{
+		"_comment": "This test vector comes from https://github.com/C2SP/CCTV/blob/5ea85644bd035c555900a2f707f7e4c31ea65ced/ed25519vectors/ed25519vectors.json",
+		"number": 0,
+		"key": "0000000000000000000000000000000000000000000000000000000000000000",
+		"sig": "00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+		"msg": "ed25519vectors 3",
+		"flags": [
+			"low_order_R",
+			"low_order_A",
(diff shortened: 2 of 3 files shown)


rozbb:
I think there was just never a torsion-freeness check. See [original commit](https://github.com/dalek-cryptography/ed25519-dalek/pull/98/files#diff-68515e4c74805dc33a55c831647e538e25679b76d71b93485d827490f58aa1a6R329) and [further discusision](https://github.com/dalek-cryptography/ed25519-dalek/issues/20).

If this is the case, we have a few options:
1. Tighten up `verify_strict()` to be the thing we claimed
2. Remove `verify_strict()` entirely if we can't figure out why people want it
3. Leave `verify_strict()` alone and just update the docs
4. Write a `verify_extra_strict()` method that checks for torsion-freeness

tarcieri:
Are there any use cases for `verify_strict` other than consensus-critical applications where signatures are committed in a Merkle tree?

I don't think `verify_strict` is helpful in these cases anyway because there aren't a well-defined, interoperable set of rules for strict verification. And especially if you're talking about a `verify_extra_strict`, the construction is becoming even more bespoke.

For the application of signed data in a Merkle tree, I think any of the following solutions are better:

1. Hoist the signatures out of the Merkle tree, as in certificate transparency's SCTs
2. Move the signatures out-of-band from the signed data, as in SegWit
3. Use a VRF instead of a signature, and commit to the VRF output (sans proof) as in KT

rozbb:
I think I agree with you here. Can we land this PR and then do something about `verify_strict` in a separate one?

tarcieri:
Sure, seems good

Issue #254, opened 2023-01-02
SecretKey creation


SergeStrashko:
[Here](https://github.com/dalek-cryptography/ed25519-dalek/blob/7529d65506147b6cb24ca6d8f4fc062cac33b395/src/secret.rs#L110)
key initialization implicitly requires that passed slice has exact length of 32 and the only case when this function returns error is length mismatch.

```rust
    pub fn from_bytes(bytes: &[u8]) -> Result<SecretKey, SignatureError> {
        if bytes.len() != SECRET_KEY_LENGTH {
            return Err(InternalError::BytesLengthError {
                name: "SecretKey",
                length: SECRET_KEY_LENGTH,
            }.into());
        }
        let mut bits: [u8; 32] = [0u8; 32];
        bits.copy_from_slice(&bytes[..32]);

        Ok(SecretKey(bits))
    }
```

shouldn't it be just (or can we add; I can work on it)
following implementation? 

```rust
impl From<[u8; SECRET_KEY_LENGTH]> for SecretKey {
     fn from(key: [u8; SECRET_KEY_LENGTH]) -> Self {
         SecretKey(key)
     }
}
```
It could have few benefits IMHO.

1. no need to deal with `Result`
2.  no extra operations copying from source slice

PR #255, opened 2023-01-02
add From<[u8; 32]> for SecretKey

SergeStrashko:
Proposed implementation of #254

### DIFF
--- a/src/secret.rs
+++ b/src/secret.rs
@@ -180,6 +180,13 @@ impl SecretKey {
     }
 }
 
+
+impl From<[u8; SECRET_KEY_LENGTH]> for SecretKey {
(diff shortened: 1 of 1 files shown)


tarcieri:
You need to target the `release/2.0` branch, where `SecretKey` is already a type alias for `[u8; 32]`.

@rozbb perhaps we should merge `release/2.0` into `main` and bump the version to `2.0.0-pre` ala what we did for curve25519-dalek

SergeStrashko:
Noted. Thank you.

tarcieri:
This API permits infallibly initializing from a 32-byte `SecretKey`: https://github.com/dalek-cryptography/ed25519-dalek/blob/e2ed313/src/signing.rs#L92-L98

There's a `TryFrom<&[u8]>` impl as well.

PR #260, opened 2023-01-08
Use `curve25519-dalek` from git; check in Cargo.lock

tarcieri:
Updates to the latest upstream changes in `curve25519-dalek`, including using the new `EdwardsPoint::mul_base` API.

To keep the build deterministic, this also checks in Cargo.lock, which pins `curve25519-dalek` to a particular git commit SHA which can be updated using `cargo update -p curve25519-dalek`.

We can potentially remove `Cargo.lock` again after a crate release.


tarcieri:
Note: I'm a big fan of checking in Cargo.lock in general, [despite the advise here](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries).

The absence of a deterministic build makes build failures like this one from #251 difficult to reason about, because it's hard to answer the question of what changed since it's outside of source control: https://github.com/dalek-cryptography/ed25519-dalek/actions/runs/3868407750/jobs/6593824474

It also means you can't reliably bisect past versions, because they may break due to environmental changes in crates.io like yanking.

pinkforest (approved):
I was one of the people who used to think .lock is pointless for libs too but I would recommend it too now having seen the light esp for tracking testing and understand what was used (implicitly) given time in history.

rozbb:
I think it's reasonable to do it as a temporary fix, but I'm not sure I agree with shipping the `Cargo.lock` file in a release. Another solution to our build problem would be to make a dalek monorepo.

Accepting for now.

pinkforest:
Also when it's monorepo the lock can still loiter around workspace root without polluting library crates themselves which can also have test-binary crate used for testing that can leverage the .lock

PR #261, opened 2023-01-08
Bump `ed25519` crate to v2.0.0-rc.1

tarcieri:
This transitively updates `signature` to v2.0.0-rc.1, which contains RNG API changes which don't impact `ed25519-dalek`, as it doesn't use or enable the RNG API.


tarcieri:
I'll have `rc.2` out tomorrow which uses `signature` v2.0.0 (final release)

PR #262, opened 2023-01-09
Make `rand_core` optional

pinkforest:
As documented here via DOC PR #241 :
- https://github.com/pinkforest/ed25519-dalek/tree/docs-docsrs#Feature-flags

I have further proposal how to re-organise the examples to clear out lib .rs but I'll do it in separate PR

~~Depends on to get rid of `rand` here on `batch` feature:~~
- #256 - This PR turned into fix-up PR as some aligned fixes were there already


Cargo.toml
```diff
@@ -60,6 +61,7 @@ batch = ["alloc", "merlin", "rand_core"]
 legacy_compatibility = []
 pkcs8 = ["ed25519/pkcs8"]
 pem = ["alloc", "ed25519/pem", "pkcs8"]
+rand_core = ["dep:rand_core"]
```
pinkforest, line 64: It is best to keep iterating explicitly all the features in Cargo.toml as it's easier to understand for downstream

src/signing.rs
```diff
@@ -183,7 +184,10 @@ impl SigningKey {
     /// The standard hash function used for most ed25519 libraries is SHA-512,
     /// which is available with `use sha2::Sha512` as in the example above.
     /// Other suitable hash functions include Keccak-512 and Blake2b-512.
-    #[cfg(feature = "rand_core")]
+    ///
+    /// # Features
+    ///
+    /// Requires optional feature `rand_core` activated
```
robjtede, line 190: the crate is set up to allow this:

```suggestion
#[cfg_attr(docsrs, doc(cfg(feature = "rand_core")))]
#[cfg(feature = "rand_core")]
```

tarcieri: Wouldn't it be good to move to `doc_auto_cfg` instead?

robjtede: Maybe; my personal view is to wait for this to be stable before using it given what I'm reading on the tracking issue. Either way, using the built-in attributes should be preferred over doc comments.

tarcieri: Can you speak to those issues in particular? Reading the list linked here:

https://github.com/rust-lang/rust/pull/100883#issuecomment-1317684866

...it seems like if `doc_auto_cfg` generates the attributes correctly, barring some other nightly regression (which won't retroactively impact previous successful docs.rs builds) it should just be a matter of migrating to `#![doc(auto_cfg)]` when the feature stabilizes.

pinkforest: The gate is at the top not at the fn. The documentation documents the feature-gate that is all.

pinkforest: we setup rand_core in the features for the docsrs in the metadata. the docsrs changes are in another PR

src/signing.rs
```diff
@@ -153,6 +153,7 @@ impl SigningKey {
         self.verifying_key
     }
 
+    #[cfg(any(test, feature = "rand_core"))]
```
pinkforest, line 156: The gate is already set here for generate() and it ensures we don't need to gate doctests in teh doc itself
We don't gate with docsrs since we setup the features for docsrs separately in the docs PR
This aligns with everything else - the gates sit above comments always.

tarcieri: > the gates sit above comments always.

FWIW I've found that style to be... unusual to me.

I'm used to a "comments, attributes, code" style which is used by e.g. rustc

pinkforest: It's easier to read imo and we have it elsewhere like that ? we could always change that

pinkforest: It's also that we have some attr's middle of doctests.. I'm pondering about the doctests

tarcieri: My vote would be to change it as I haven't seen it used anywhere else. 

Unfortunately I don't see the "comments, attributes, code" ordering actually prescribed in any style guide, but it's pretty consistently that way in every project I've seen maintained by the Rust team

pinkforest: Your wish shall be granted. psst we also needed it for test as there are some test fn's that use generate internally.

pinkforest: fwiw - docsrs is / has been running nightly and we would only set the unstable flag with docsrs anyways ?
`#![cfg_attr(docsrs, feature(doc_auto_cfg))]`
We already use doc_auto_cfg in curve25519
https://github.com/dalek-cryptography/curve25519-dalek/blob/main/src/lib.rs#L14

tarcieri: I've migrated several @RustCrypto projects to using it, and it seems to be working fine to me. I've checked the output versus the manual annotations and haven't found any discrepancies besides things I missed in manual annotations

pinkforest: Yah, I took the manual annotation out. Thanks for picking this.

PR #264, opened 2023-01-09
fixed small typo errors

0xflotus:


rozbb:
Thank you! Actually, we're planning a major overhaul on our statements about `verify_strict`, so all these paragraphs are gonna get deleted in the end 😬

PR #268, opened 2023-01-17
Make `digest` optional

pinkforest:
The crate `digest` isn't yet stable but this is exposed it via the public API from curve25519_dalek `digest` re-export.

This PR makes the public API use optional via an optional digest feature.

This is documented in the DOC PR: https://github.com/dalek-cryptography/ed25519-dalek/pull/241 ([Render](https://github.com/pinkforest/ed25519-dalek/tree/docs-docsrs))

**Affected API - now behind cfg(feature = "digest")**
- `ed25519_dalek::Digest` - is re-exported via public API where _prehashed fn's reflect D:gating the trait
- `SigningKey::sign_prehashed(D: prehashed_message, ..)`
- `SigningKey::verify_prehashed(D: prehahed_message, ..)` - proxy of `VerifyingKey::verify_prehashed`
- `VerifyingKey::verify_prehashed(D: prehashed_message, ..)`
- `VerifyingKey::verify_prehashed_strict(D: prehashed_message, ..)`

**Questions**
- `ed25519_dalek::Sha512` - Was this supposed to be re-exported via public API and to integration tested as doctests seem to refer use `ed25519_dalek::Sha512` meanwhile tests/ed25519 uses `sha2::Sha512` ? :woman_shrugging: 
- `ed25519_dalek::Sha512` - remove re-export - was re-exported via `verifying.rs` via `pub use verifying::*`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -262,6 +262,7 @@ mod signature;
 mod signing;
 mod verifying;
 
+#[cfg(feature = "digest")]
 pub use curve25519_dalek::digest::Digest;
 
 #[cfg(feature = "batch")]
--- a/src/signing.rs
+++ b/src/signing.rs
@@ -24,6 +24,7 @@ use serde_bytes::{ByteBuf as SerdeByteBuf, Bytes as SerdeBytes};
 
 use sha2::Sha512;
 
+#[cfg(feature = "digest")]
 use curve25519_dalek::digest::generic_array::typenum::U64;
 use curve25519_dalek::digest::Digest;
 use curve25519_dalek::edwards::CompressedEdwardsY;
@@ -208,12 +209,15 @@ impl SigningKey {
     ///
     /// # Examples
     ///
-    #[cfg_attr(feature = "rand_core", doc = "```")]
-    #[cfg_attr(not(feature = "rand_core"), doc = "```ignore")]
+    #[cfg_attr(all(feature = "rand_core", feature = "digest"), doc = "```")]
+    #[cfg_attr(
+        any(not(feature = "rand_core"), not(feature = "digest")),
+        doc = "```ignore"
+    )]
     /// use ed25519_dalek::Digest;
     /// use ed25519_dalek::SigningKey;
-    /// use ed25519_dalek::Sha512;
     /// use ed25519_dalek::Signature;
+    /// use sha2::Sha512;
     /// use rand::rngs::OsRng;
     ///
     /// # #[cfg(feature = "std")]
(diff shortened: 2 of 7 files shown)


src/signing.rs
```diff
@@ -208,8 +209,11 @@ impl SigningKey {
...
-    #[cfg_attr(feature = "rand_core", doc = "```")]
-    #[cfg_attr(not(feature = "rand_core"), doc = "```ignore")]
+    #[cfg_attr(all(feature = "rand_core", feature = "digest"), doc = "```")]
+    #[cfg_attr(
+        any(not(feature = "rand_core"), not(feature = "digest")),
+        doc = "```ignore"
+    )]
     /// use ed25519_dalek::Digest;
     /// use ed25519_dalek::SigningKey;
     /// use ed25519_dalek::Sha512;
```
pinkforest, line 219: So are we supposed to re-export Sha512 ? If not may neet to remove this here.

src/lib.rs
```diff
@@ -262,7 +262,10 @@ mod signature;
 mod signing;
 mod verifying;
 
+#[cfg(feature = "digest")]
 pub use curve25519_dalek::digest::Digest;
+#[cfg(feature = "digest")]
+pub use sha2::Sha512;
```
pinkforest, line 268: If we are not supposed to re-export Sha512 then I need to remove Sha512 re-export here.
If we are supposed to re-export Sha512 then I need to adjust some integration tests to refer our re-export

src/signing.rs
tarcieri: How were these passing before?

I think it's ok to gate a re-export behind the `digest` feature.

rozbb: Why were we exporting `Sha512` at all? I'm not entirely opposed, it just seems like an odd choice. People should probably bring their own digest function. The fact we use Sha512 under the hood is more an implementation detail.

rozbb:
Looks great! Just one loose thread above before we merge.

src/signing.rs
pinkforest: It's indeed a mystery how these were passing before as Sha512 was never exported.
It's also complicated since the Digest'd output type comes essentially from Sha512 which we require as input and there is whole lot of back/forward related coupling related to that.
I would say that makes it feasible to export for compatibility reasons as your dependency tree might have incompat types beween versions when Digest is used and using the exported Sha512 gives the right Digest input.
That said this wasn't exported before but I think if we are using it in the API gating it's may usually be best to export.

src/lib.rs
pinkforest, line 268: We should probably change this later to re-export this from curve like we export Digest from

src/signing.rs
pinkforest: The plot thickens.. the Sha512 was there always.. guess why? src/lib.rs

```rust
pub use crate::constants::*;
pub use crate::errors::*;
pub use crate::signing::*;
pub use crate::verifying::*;

```... as this also exported Sha512 guess from where ? :D

pinkforest: spoiler: it came from verifying.rs which did pub re-export of Sha512 that then got re-exported at lib.rs

Issue #267, opened 2023-01-17
Validation criteria: `verify_strict` might be accepting non-canonical `R`


rozbb:
It [appears](https://github.com/dalek-cryptography/ed25519-dalek/blob/6d9bbd323edfce04f600427571e90afd86f52939/src/verifying.rs#L307) that `verify_strict` might permit [non-canonical `R`](https://hdevalence.ca/blog/2020-10-04-its-25519am), while `verify` does not. Notice the difference in the code is that `verify_strict` does a point comparison, while `verify` does a byte comparison.

This is pretty subtle, and I think it's not detected by our validation tests because the validation tests never do a non-canonical `R` that isn't also low-order. Three things I wanna do:
1. Confirm that this behavior is happening
2. Update `verify_strict` to reject non-canonical encoding
3. Write a regression test similar to what [zebra has](https://github.com/ZcashFoundation/ed25519-zebra/blob/612e51af2e7eaa228f9a08ad38b0b0660e510d9e/tests/util/mod.rs#L81-L155)

This is technically breaking, but I consider it a bugfix rather than a validation change. Nobody would reasonably expect `verify_strict` to be more permissive than `verify` on any axis.

rozbb:
Ok I think this is a non-issue. Yes, it may be the case that `verify_strict` accepts prime-order non-canonical `R`, but the reason it's not an issue is because it's extremely difficult to generate a signature with such an `R`. That's because the ability to produce such a signature is the same as proving knowledge of of `dlog(R)`, which is probably hard.

My proposed solution: let's just make this explicit, and do verification by byte comparison in all `verify_*` functions. This will simplify the code and only change behavior in a computationally infeasible edge case.

cc @tarcieri

tarcieri:
> My proposed solution: let's just make this explicit, and do verification by byte comparison in all verify_* functions. 

Agree with this, and I wonder if you could potentially wrap it up a little more, perhaps a function that calls your proposed `recompute_r` function and compares it with an expected `R` in a single place which can be reused?

rozbb:
My thought was to have `recompute_r` just return the compressed form, and then all the callers have no choice but to do a byte comparison. This looks nice enough imo, since everything all the verify functions look like `expected_R == computed_R`.

But #251 needs to land before that happens.

tarcieri:
ed25519-dalek has moved to: https://github.com/dalek-cryptography/curve25519-dalek/tree/main/ed25519-dalek

[...]

PR #269, opened 2023-01-20
Do byte comparison in all `verify_*` functions

rozbb:
This is just cleanup. As explained in #267, this doesn't change validation criteria.

This also removes manual clamping from this crate and uses `curve25519_dalek::Scalar::from_bits_clamped` instead, now that https://github.com/dalek-cryptography/curve25519-dalek/pull/498 has landed.

### DIFF
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -184,7 +184,8 @@ impl TryFrom<&ed25519::Signature> for InternalSignature {
 impl From<InternalSignature> for ed25519::Signature {
     #[allow(clippy::unwrap_used)]
     fn from(sig: InternalSignature) -> ed25519::Signature {
-        // This function is actually infallible but it's marked as fallible
+        // This function only fails if the s half of the parsed input exceeds the scalar modulus.
+        // Since the bytes are coming straight from a Scalar, this is impossible.
         ed25519::Signature::from_bytes(&sig.as_bytes()).unwrap()
     }
 }
--- a/src/signature.rs
+++ b/src/signature.rs
@@ -162,6 +162,7 @@ impl InternalSignature {
     /// only checking the most significant three bits.  (See also the
     /// documentation for [`crate::VerifyingKey::verify_strict`].)
     #[inline]
+    #[allow(clippy::unwrap_used)]
     pub fn from_bytes(bytes: &[u8; SIGNATURE_LENGTH]) -> Result<InternalSignature, SignatureError> {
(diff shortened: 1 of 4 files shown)


src/signing.rs
```diff
@@ -671,24 +671,13 @@ impl Drop for ExpandedSecretKey {
...
-        hash.copy_from_slice(h.finalize().as_slice());
-
-        lower.copy_from_slice(&hash[00..32]);
-        upper.copy_from_slice(&hash[32..64]);
-
-        lower[0] &= 248;
-        lower[31] &= 63;
-        lower[31] |= 64;
+        let hash = Sha512::default().chain_update(secret_key).finalize();
+        let (lower, upper) = hash.split_at(32);
```
tarcieri, line 675: Maybe leave a comment to use `array_chunks` when stable to get rid of the `unwrap`s below?

(might also be good to add a `clippy::unwrap_used` lint too)

tarcieri (approved):

src/signing.rs
rozbb: What do you mean add the lint? As in make it a global Deny, and then do an explicit Allow in this function?

tarcieri: Yeah

src/signature.rs
```diff
@@ -181,7 +182,9 @@ impl TryFrom<&ed25519::Signature> for InternalSignature {
 }
 
 impl From<InternalSignature> for ed25519::Signature {
+    #[allow(clippy::unwrap_used)]
     fn from(sig: InternalSignature) -> ed25519::Signature {
+        // This function is actually infallible but it's marked as fallible
```
rozbb, line 187: Just making sure. Is this true? [Source](https://github.com/RustCrypto/signatures/blob/0bb3f07a0677b4c87a1b580c81eac68bbfc69f1c/ed25519/src/lib.rs#L304-L311)

tarcieri: The conversion won't ever fail because a serialized `InternalSignature` will never violate `ed25519::Signature`'s invariant that `s` does not overflow the curve's order.

However, if it were to, `ed25519::Signature::from_bytes` will fail.

rozbb:
Note: `release/2.0` build is failing because `ed25519 v2.0` was yanked. Rather than modifying `Cargo.toml` I'm just gonna wait until `2.0.1` comes out.

PR #270, opened 2023-01-21
Impl `signature::Digest*` traits for Ed25519ph

tarcieri:
Adds the following trait impls:

- impl DigestSigner for SigningKey
- impl DigestVerifier for VerifyingKey

These traits can be used to create and verify Ed25519 signatures, thunking to `SigningKey::sign_prehashed` and
`VerifyingKey::verify_prehashed` respectively.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -301,6 +301,7 @@ dependencies = [
  "serde_bytes",
  "serde_json",
  "sha2",
+ "signature",
  "toml",
  "zeroize",
 ]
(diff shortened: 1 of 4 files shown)


rozbb:
Would it show up in docs if you made a comment like "this is equivalent to doing `verify_prehashed`/`sign_prehashed` with `context = None`"? Would be good to make it clear,  since it's ambiguous otherwise.

tarcieri:
Yes, I can add doc comments that show up in rustdoc

PR #271, opened 2023-01-21
CI: test `cargo doc` build

tarcieri:
Ensure it's free of warnings

### DIFF
--- a/src/signing.rs
+++ b/src/signing.rs
@@ -43,6 +43,7 @@ use crate::constants::*;
 use crate::errors::*;
 use crate::signature::*;
 use crate::verifying::*;
+use crate::Signature;
 
 /// ed25519 secret key as defined in [RFC8032 § 5.1.5]:
 ///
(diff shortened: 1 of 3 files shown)


src/signing.rs
```diff
@@ -40,6 +40,7 @@ use crate::constants::*;
 use crate::errors::*;
 use crate::signature::*;
 use crate::verifying::*;
+use crate::Signature;
```
tarcieri, line 43: This fixes the following broken link in the documentation:

```
warning: unresolved link to `Signature`
   --> src/signing.rs:320:5
    |
320 | /     /// Verify a `signature` on a `prehashed_message` using the Ed25519ph algorithm.
321 | |     ///
322 | |     /// # Inputs
323 | |     ///
...   |
382 | |     ///
383 | |     /// [rfc8032]: https://tools.ietf.org/html/rfc8032#section-5.1
    | |__________________________________________________________________^
    |
    = note: the link appears in this line:

            * `signature` is a purported Ed25519ph [`Signature`] on the `prehashed_message`.
                                                    ^^^^^^^^^^^
```

PR #275, opened 2023-01-27
README changes for 2.0

rozbb:
[Rendered](https://github.com/dalek-cryptography/ed25519-dalek/blob/b635fc717e2ebb6a9dedbe6413d7ede9cdd14d56/README.md)

Other changes:
* Feature-gated `Context` and `with_context` behind `digest`. No reason to have those if you can't use them.
* Bumped the version number to 2.0.0-pre.0
* Added a `CONTRIBUTING.md` and updated authors list in `Cargo.toml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-# ed25519-dalek [![](https://img.shields.io/crates/v/ed25519-dalek.svg)](https://crates.io/crates/ed25519-dalek) [![](https://docs.rs/ed25519-dalek/badge.svg)](https://docs.rs/ed25519-dalek) [![](https://travis-ci.org/dalek-cryptography/ed25519-dalek.svg?branch=master)](https://travis-ci.org/dalek-cryptography/ed25519-dalek?branch=master)
+# ed25519-dalek [![](https://img.shields.io/crates/v/ed25519-dalek.svg)](https://crates.io/crates/ed25519-dalek) [![](https://docs.rs/ed25519-dalek/badge.svg)](https://docs.rs/ed25519-dalek) [![Rust](https://github.com/dalek-cryptography/curve25519-dalek/actions/workflows/rust.yml/badge.svg?branch=main)](https://github.com/dalek-cryptography/curve25519-dalek/actions/workflows/rust.yml)
 
 Fast and efficient Rust implementation of ed25519 key generation, signing, and
-verification in Rust.
+verification.
 
 # Use
 
@@ -23,12 +23,12 @@ ed25519-dalek = "2.0.0-pre.0"
 
 # Feature Flags
 
-This crate is `#[no_std]` compatible with `default-features = false`
+This crate is `#[no_std]` compatible with `default-features = false`.
 
 | Feature                | Default? | Description |
 | :---                   | :---     | :---        |
-| `alloc`                | ✓        | Enables features that require dynamic heap allocation |
-| `std`                  | ✓        | Implements `std::error::Error` for `SignatureError` |
+| `alloc`                | ✓        | When `pkcs8` is enabled, implements `EncodePrivateKey`/`EncodePublicKey` for `SigningKey`/`VerifyingKey`, respectively. |
+| `std`                  | ✓        | Implements `std::error::Error` for `SignatureError`. Also enables `alloc`. |
 | `zeroize`              | ✓        | Implements `Zeroize` and `ZeroizeOnDrop` for `SigningKey` |
 | `rand_core`            |          | Enables `SigningKey::generate` |
 | `batch`                |          | Enables `verify_batch` for verifying many signatures quickly. Also enables `rand_core`. |
@@ -59,10 +59,9 @@ See [CHANGELOG.md](CHANGELOG.md) for a list of changes made in past version of t
 
 Documentation is available [here](https://docs.rs/ed25519-dalek).
 
-# Policies
+# Compatibility Policies
 
-All on-by-default features of this library are covered by semantic versioning (SemVer)
-
+All on-by-default features of this library are covered by [semantic versioning](https://semver.org/spec/v2.0.0.html) (SemVer).
 SemVer exemptions are outlined below for MSRV and public API.
 
 ## Minimum Supported Rust Version
@@ -72,11 +71,11 @@ SemVer exemptions are outlined below for MSRV and public API.
 | 2.x      | 1.60   |
 | 1.x      | 1.41   |
 
-MSRV changes will be accompanied by a minor version bump.
+From 2.x and on, MSRV changes will be accompanied by a minor version bump.
 
 ## Public API SemVer Exemptions
 
-Breaking changes to SemVer exempted components affecting the public API will be accompanied by some version bump.
+Breaking changes to SemVer-exempted components affecting the public API will be accompanied by some version bump.
 
 Below are the specific policies:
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -19,28 +19,32 @@ This crate is `#[no_std]` compatible with `default-features = false`
...
-| `pem`                  |          | PEM Support |
-| `rand_core`            |          | TODO        |
+| `std`                  | ✓        | Implements `std::error::Error` for `SignatureError` |
+| `zeroize`              | ✓        | Implements `Zeroize` and `ZeroizeOnDrop` for `SigningKey` |
+| `rand_core`            |          | Enables `SigningKey::generate` |
+| `batch`                |          | Enables `verify_batch` for verifying many signatures quickly. Also enables `rand_core`. |
+| `digest`               |          | Enables `Context`, `SigningKey::{with_context, sign_prehashed}` and `VerifyingKey::{with_context, verify_prehashed, verify_prehashed_strict}` for Ed25519ph prehashed signatures |
+| `asm`                  |          | Enables assembly optimizations in the SHA-512 compression functions |
+| `pkcs8`                |          | Enables [PKCS#8](https://en.wikipedia.org/wiki/PKCS_8) serialization/deserialization for `SigningKey` and `VerifyingKey` |
+| `pem`                  |          | TODO |
```
rozbb, line 29: @tarcieri  can you fill this and `pkcs8` out? I'm pretty confused actually. How is `pem` different from PKCS#8? Isn't PEM a container format within the PKCS#8 spec? What does it contain? What are the other things in PKCS#8?

README.md
```diff
@@ -139,79 +145,44 @@ Please see the [curve25519_dalek backend documentation](https://docs.rs/curve255
...
+
+## Batch verification
+
+As mentioned above, weak public keys can be used to produce signatures for unknown messages with high probability. This means that sometimes a weak forgery attempt will fail. In fact, it can fail up to 7/8 of the time. If you call `verify()` twice on the same failed forgery, it will return an error both times, as expected. However, if you call `verify_batch()` twice on two distinct otherwise-valid batches, both of which contain the failed forgery, there's a 21% chance that one fails and the other succeeds.
+
+Why is this? It's because `verify_batch()` does not do the weak key testing of `verify_strict()`, and it multiplies each verification equation by some random coefficient. If the failed forgery gets multiplied by 8, then the weak key (which is a low-order point) becomes 0, and the verification equation on the attempted forgery will succeed.
+
+Since `verify_batch()` is intended to be high-throughput, we think it's best not to put weak key checks in it. If you want to prevent weird behavior due to weak public keys in your batches,
+TODO
+you should call `VerifyingKey::is_weak` on the inputs in advance.
```
rozbb, line 186: @tarcieri I think this would be a good mitigation to have. Shall we merge something like #188? I don't think we need the torsion check though

rozbb:
These are most of the changes I wanted to make to the README. I might add some tweaks here and there tho

README.md
pinkforest: PEM "Privacy Enchanced Email" (yes confusing) is a text file format .. like Base64 wrapping/formatting of binary X.509 certificates e.g. PEM is the --BEGIN ASCII blocks that simply contain the Base64 formatting of the PKCS#8
The origin was from the early internet as e-mail in it's 7bit form wasn't really designed to carry binary data and it was encoded and wrapped (the --BEGIN block) in as Base64 text formatting that allowed itself to be carried in 7bit SMTP - https://en.wikipedia.org/wiki/Base64
PKCS#8 in turn is the binary encoded container under it which may contain passphrase etc - it's just ASN.1/BER (or was it DER) whilst the PEM is the text Base64 wrapping on top of it
PKCS#8 ASN.1 schema is here - https://datatracker.ietf.org/doc/html/rfc5208#page-5

README.md
```diff
@@ -71,9 +75,11 @@ Below are the specific policies:
 | :---     | :---                    | :---   |
 | 2.x      | Dependencies `digest`, `pkcs8` and `rand_core` | Minor SemVer bump |
 
+TODO @tarcieri what about PEM?
```
pinkforest, line 78: We only lists the crates / dependencies but not features as such here -

`pem` as a feature comes via `pkcs8` (unstable) crate via optional feature of `ed25519` which itself is stable on all-default-on -

Since it's not a crate or dependency itself unlike pkcs8 - theoretically we don't need to mention `pem` here, just `pkcs8` given it's optional dependency of `ed25519` thus exempted.

Given anyone who is using `pem` will inherently use `pkcs8` I don't see much problem not mentioning `pem` here ?

README.md
```diff
@@ -19,28 +19,32 @@ This crate is `#[no_std]` compatible with `default-features = false`
...
-| `legacy_compatibility` |          | See: A Note on Signature Malleability |
-| `pkcs8`                |          | PKCS#8 Support |
-| `pem`                  |          | PEM Support |
-| `rand_core`            |          | TODO        |
+| `std`                  | ✓        | Implements `std::error::Error` for `SignatureError` |
+| `zeroize`              | ✓        | Implements `Zeroize` and `ZeroizeOnDrop` for `SigningKey` |
+| `rand_core`            |          | Enables `SigningKey::generate` |
+| `batch`                |          | Enables `verify_batch` for verifying many signatures quickly. Also enables `rand_core`. |
+| `digest`               |          | Enables `Context`, `SigningKey::{with_context, sign_prehashed}` and `VerifyingKey::{with_context, verify_prehashed, verify_prehashed_strict}` for Ed25519ph prehashed signatures |
+| `asm`                  |          | Enables assembly optimizations in the SHA-512 compression functions |
```
pinkforest, line 27: I re-ordered this alphabetically between default/optionals - just wonderign what is the ordering here so they are easily found when looked up ? also legacy_compatibility last ?

README.md
```diff
@@ -139,79 +145,44 @@ Please see the [curve25519_dalek backend documentation](https://docs.rs/curve255
...
+
+`ed25519-dalek` is not a malleable signature scheme.
+
+Some other Ed25519 implementations are malleable, though, such as [libsodium with `ED25519_COMPAT` enabled](https://github.com/jedisct1/libsodium/blob/24211d370a9335373f0715664271dfe203c7c2cd/src/libsodium/crypto_sign/ed25519/ref10/open.c#L30), [ed25519-donna](https://github.com/floodyberry/ed25519-donna/blob/8757bd4cd209cb032853ece0ce413f122eef212c/ed25519.c#L100), [NaCl's ref10 impl](https://github.com/floodyberry/ed25519-donna/blob/8757bd4cd209cb032853ece0ce413f122eef212c/fuzz/ed25519-ref10.c#L4627), and probably a lot more.
+
+If you need to interoperate with such implementations and accept otherwise invalid signatures, you can enable the `legacy_compatibility` flag. **Do not enable `legacy_compatibility`** if you don't have to, because it will make your signatures malleable.
+
+Note: [CIRCL](https://github.com/cloudflare/circl/blob/fa6e0cca79a443d7be18ed241e779adf9ed2a301/sign/ed25519/ed25519.go#L358) has no scalar range check at all. We do not have a feature flag for interoperating with the larger set of RFC-disallowed signatures that CIRCL accepts.
+
+## Weakkey Forgery and `verify_strict()`
```
pinkforest, line 170:
```suggestion
## Weak key Forgery and `verify_strict()`
```

README.md
rozbb: I think alphabetical is fine. The previous ordering was thematic, where the themes are things only I had in my head :)

README.md
```diff
@@ -1,55 +1,67 @@
-# ed25519-dalek [![](https://img.shields.io/crates/v/ed25519-dalek.svg)](https://crates.io/crates/ed25519-dalek) [![](https://docs.rs/ed25519-dalek/badge.svg)](https://docs.rs/ed25519-dalek) [![](https://travis-ci.org/dalek-cryptography/ed25519-dalek.svg?branch=master)](https://travis-ci.org/dalek-cryptography/ed25519-dalek?branch=master)
+# ed25519-dalek [![](https://img.shields.io/crates/v/ed25519-dalek.svg)](https://crates.io/crates/ed25519-dalek) [![](https://docs.rs/ed25519-dalek/badge.svg)](https://docs.rs/ed25519-dalek) [![Rust](https://github.com/dalek-cryptography/curve25519-dalek/actions/workflows/rust.yml/badge.svg?branch=main)](https://github.com/dalek-cryptography/curve25519-dalek/actions/workflows/rust.yml)
```
pinkforest, line 1:
```suggestion
# ed25519-dalek [![](https://img.shields.io/crates/v/ed25519-dalek.svg)](https://crates.io/crates/ed25519-dalek) [![](https://docs.rs/ed25519-dalek/badge.svg)](https://docs.rs/ed25519-dalek) [![Rust](https://github.com/dalek-cryptography/ed25519-dalek/actions/workflows/rust.yml/badge.svg?branch=main)](https://github.com/dalek-cryptography/ed25519-dalek/actions/workflows/rust.yml)

some sneaky curve25519 references there :)
```

README.md
tarcieri, line 38:
```suggestion
| `pem`                  |          | Enables PEM serialization support for PKCS#8 private keys and SPKI public keys |
```

As to why they're separate features: `pkcs8` works in heapless environments. `pem` (currently) requires `alloc`.

Streaming encoding/decoding is almost working on the latest `der`/`pkcs8` prereleases though, which would eliminate that restriction so they don't need to be separate features.

With weak feature activation, it would also be possible to make `pkcs8` + `alloc` automatically enable `pem`.

PR #279, opened 2023-02-04
Bump curve25519 dep to rc.1

rozbb:
I also realized that checking in `Cargo.lock` breaks the MSRV check, since it no longer regens the lockfile. The MSRV check now removes the lockfile before doing anything.


rozbb:
This is very small but I can't commit directly to `main`

tarcieri (approved):

PR #283, opened 2023-02-24
Remove invalid input fields from CI action for Rust setup

striezel:
Neither `override` nor `profile` are valid inputs for the `dtolnay/rust-toolchain` action. It always uses the minimal profile anyways.

This PR gets rid of the warning

> Unexpected input(s) 'override', 'profile', valid inputs are ['toolchain', 'targets', 'target', 'components']

as seen in https://github.com/dalek-cryptography/ed25519-dalek/actions/runs/4092160926.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -105,6 +105,4 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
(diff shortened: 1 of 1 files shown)


robjtede (approved):

rozbb (approved):
Good catch. Thank you!

PR #285, opened 2023-03-05
Bump `ed25519` to v2.2; `pkcs8` to v0.10

tarcieri:
The `ed25519` v2.2.0 crate bumps the `pkcs8` dependency to v0.10.

This updates `ed25519` to the latest version and updates the PKCS#8 support to use the new API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ features = ["batch", "pkcs8"]
 
 [dependencies]
 curve25519-dalek = { version = "=4.0.0-rc.1", default-features = false, features = ["digest"] }
-ed25519 = { version = "2.1", default-features = false }
+ed25519 = { version = ">=2.2, <2.3", default-features = false }
 signature = { version = ">=2.0, <2.1", optional = true, default-features = false }
 sha2 = { version = "0.10", default-features = false }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,9 +135,9 @@ dependencies = [
 
 [[package]]
 name = "const-oid"
-version = "0.9.1"
+version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cec318a675afcb6a1ea1d4340e2d377e56e47c266f28043ceccbf4412ddfdd3b"
+checksum = "520fbf3c07483f94e3e3ca9d0cfd913d7718ef2483d2cfd91c0d9e91474ab913"
 
 [[package]]
 name = "cpufeatures"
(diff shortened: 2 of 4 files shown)


tarcieri:
Note: MSRV of PKCS#8 support becomes 1.65 with this update. I will need to update the CI config accordingly.

Cargo.toml
```diff
@@ -26,7 +26,7 @@ features = ["batch", "pkcs8"]
 
 [dependencies]
 curve25519-dalek = { version = "=4.0.0-rc.1", default-features = false, features = ["digest"] }
-ed25519 = { version = "2.1", default-features = false }
+ed25519 = { version = ">=2.2, <2.3", default-features = false }
```
robjtede, line 29: This kind of version range has been pretty detrimental to downstream users in the past. Was v2.2 _technically_ breaking?

tarcieri: Look at the PR. There are several breaking changes to PKCS#8 support.

It's following the stated policy here:

https://github.com/RustCrypto/signatures/tree/master/ed25519#semver-policy

> The `pkcs8` module is exempted as it uses a pre-1.0 dependency, however, breaking changes to this module will be accompanied by a minor version bump.

As well as the one here:

https://github.com/dalek-cryptography/ed25519-dalek#public-api-semver-exemptions

robjtede: Ok got it. Why not specify the more future-friendly "2.2" dev here then?

tarcieri: Future minor releases may make further breaking changes to the `pkcs8` crate.

Once the releases are out and it's known they don't make breaking changes, the requirement can be relaxed.

mrghosti3: Anyway, I wanted to understand the reasoning of explicit version declaration. Sorry if my approach felt confronting or unpleasent. If this type of declaration is meant to be clear and verbose so that other people reading would instantly understand (without the context of how cargo parses versioning) then that LGTM. Otherwise just curious on why such preference to dependency version constraints instead of the given shorthand.

mrghosti3 (approved):

Cargo.toml
robjtede: That's not quite how versioning works in this case. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2f20d4fb80bb6d1e03e726af492e03dc).

tarcieri: Yeah, both `2.2` and `2.2.0` are interpreted as `>= 2.2.0, <3` under normal SemVer rules.

The use of versioning in place here is not specifically SemVer compatible, or that is to say, if you view the two links in my comment at the top of the thread the use of `pkcs8` is explicitly called out as being outside the scope of normal SemVer guarantees.

It's a compromise which lets part of the API get >1.x and get SemVer-compatible guarantees while still letting off-by-default features (which depend on pre-1.x crates) evolve at a different pace without forcing a major version bump every time.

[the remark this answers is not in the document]
tarcieri: As [answered above](https://github.com/dalek-cryptography/ed25519-dalek/pull/285#discussion_r1125763708), they mean different things

rozbb (approved):

Issue #287, opened 2023-03-07
Serde forces std in 2.0.0-pre.0


semenov-vladyslav:
Enabling `serde` feature only (without `std` feature) enables use of `std` crate which shouldn't happen. It seems due to bad dependencies in ed25519-dalek-2.0.0-pre.0.

robjtede:
Looks like serde_bytes has their std feature as default, and it's not disabled for the serde feature of this crate.

tarcieri:
ed25519-dalek has moved to: https://github.com/dalek-cryptography/curve25519-dalek/tree/main/ed25519-dalek

If you feel this issue has not yet been addressed, please request this issue be moved to that repo by leaving a comment, or open a new issue on that repo.

PR #288, opened 2023-03-07
Serde no_std compatibility

semenov-vladyslav:
This PR addresses issue #287. Tests pass with different combinations of features: "serde", "serde alloc", "serde std", "alloc".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ sha2 = { version = "0.10", default-features = false }
 merlin = { version = "3", default-features = false, optional = true }
 rand_core = { version = "0.6.4", default-features = false, optional = true }
 serde = { version = "1.0", default-features = false, optional = true }
-serde_bytes = { version = "0.11", optional = true }
+serde_bytes = { version = "0.11", default-features = false, optional = true }
 zeroize = { version = "1.5", default-features = false, optional = true }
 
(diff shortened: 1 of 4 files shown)


tarcieri:
Weird failure in the MSRV test. Wonder if it's a nightly bug in the unused import diagnostic.

@semenov-vladyslav maybe try pinning a slightly older nightly?

semenov-vladyslav:
Looks like clippy failed on unused imports. Forgot to update cfg pragma to a few other imports. Essentially, "serde" feature requires one of "std" or "alloc". Fixed that, now it should work.

pinkforest:
Looks like we didn't use `thumbv7em-none-eabihf` here that tests the target with no std avail.

We should probably add that: 
https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/rust.yml#L45-56

pinkforest:
Confirming no_std build works in main without serde:
https://github.com/dalek-cryptography/ed25519-dalek/actions/runs/4394998670/jobs/7696444822

However with serde it fails as expected:
https://github.com/dalek-cryptography/ed25519-dalek/actions/runs/4395018249/jobs/7696481185

Can we re-base this when the test PR in CI so we ensure in CI it's fixed.

tarcieri:

PR #289, opened 2023-03-12
Add no_std to CI

pinkforest:
The missing test in CI:
- https://github.com/dalek-cryptography/ed25519-dalek/issues/287
- https://github.com/dalek-cryptography/ed25519-dalek/pull/288

Testing that `std` is not required with a target that does not have it by `cargo-hack` excluding `default`, `std` and `serde`

`serde` is excluded and when this test is merged and re-based into #288 it can be tested there with the fix there for `serde`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -80,7 +80,7 @@ jobs:
           toolchain: stable
           targets: thumbv7em-none-eabi
       - run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features alloc
-      - run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features "alloc digest rand_core batch zeroize pkcs8"
+      - run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features "alloc digest rand_core batch zeroize pkcs8 serde"
 
   bench:
     name: Check that benchmarks compile
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -81,7 +81,8 @@ jobs:
           targets: thumbv7em-none-eabi
       - uses: taiki-e/install-action@cargo-hack
       # This tests both alloc and no-alloc without std
-      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features std
+      # TODO: serde pending PR#288
+      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features std,serde
 
   bench:
     name: Check that benchmarks compile
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -83,7 +83,7 @@ jobs:
       # No default features build
       - run: cargo build --target thumbv7em-none-eabi --release --no-default-features
       # TODO: serde pending PR#288
-      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features std,serde
+      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std,serde
 
   bench:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -70,6 +70,18 @@ jobs:
...
+  build-nostd:
+    name: Build on no_std target (thumbv7em-none-eabi)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
+          targets: thumbv7em-none-eabi
+      - run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features alloc
```
tarcieri, line 82: Maybe add a `--no-default-features` without `--features`?

.github/workflows/rust.yml
```diff
@@ -70,6 +70,18 @@ jobs:
...
+      - run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features "alloc digest rand_core batch zeroize pkcs8"
```
tarcieri, line 83: Should we consider `cargo hack` here?

.github/workflows/rust.yml
pinkforest: Sure good idea - though I think we may want two where 

1) The first is all features with no_std that doesn't require alloc and 
2) The second that has features requires alloc ?

So it tests both no_std without alloc and with alloc ?

Though there isn't target available for no_alloc or mechanism to make alloc unavailable so we may just put them all in one test ?

.github/workflows/rust.yml
pinkforest: Yeah we would have to cache it ? is there a github action that includes it or did dtolnay's toolchain action give it ?

pinkforest: Hmm it seems to burn a lot of CI minutes as it requires `--each-feature ` ?

We can do either that or just keep two sets of matrix 1) for alloc 2) without alloc

.github/workflows/rust.yml
tarcieri: I still think it's good to test that it builds without any features, which is at least an additional third test

pinkforest: k. I will leave serde exclude in the comment and if this is rebased in the fix it can be removed to check serde as well so we don't break the main

.github/workflows/rust.yml
pinkforest: Ok cargo hack seems to work - here is fail as expected with serde:
https://github.com/dalek-cryptography/ed25519-dalek/actions/runs/4395018249/jobs/7696481185
Will put the serde in there now and it should succeed here - can re-base and then take the serde exception off there

.github/workflows/rust.yml
pinkforest: Added so `cargo-hack` is the second where as vanilla `cargo build` `--no-default-features` is the first.

.github/workflows/rust.yml
pinkforest: Ahh it doesn't group the features correctly.. we need the features grouped so they activate each other.

pinkforest: Ahh need to put `--skip` for the ones which would be pointless alone - 

The feature-cascade should enable each other but do we want to check lone features ? 

Probably best to not complicate it me thinks and keep just simple exclude.

tarcieri: What features are pointless alone?

pinkforest: Hmm good question :sweat_smile: yeah none I think - I have to do little mental mindmap about the feat matrix's

tarcieri (approved):

PR #291, opened 2023-03-20
fix serde / no_std incompatibility

ryan-mob:
updates serde `Serialize` and `Deserialize` implementations to use a custom visitor, allowing `serde` to be used without `alloc` or `std` for embedded, and making this consistent with implementations in [curve25519-dalek](https://github.com/dalek-cryptography/curve25519-dalek/blob/a63e14f4ded078d6bf262ba0b3f47026bdd7f7c0/src/edwards.rs#L269).

as an alternative #288 (thanks @semenov-vladyslav)

@pinkforest seems like it'd be good to have some serde tests / this should go over #289?

### DIFF
--- a/src/signing.rs
+++ b/src/signing.rs
@@ -16,11 +16,7 @@ use ed25519::pkcs8;
 use rand_core::CryptoRngCore;
 
 #[cfg(feature = "serde")]
-use serde::de::Error as SerdeError;
-#[cfg(feature = "serde")]
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-#[cfg(feature = "serde")]
-use serde_bytes::{ByteBuf as SerdeByteBuf, Bytes as SerdeBytes};
 
 use sha2::Sha512;
 
@@ -634,7 +630,7 @@ impl Serialize for SigningKey {
     where
         S: Serializer,
     {
-        SerdeBytes::new(&self.secret_key).serialize(serializer)
+        serializer.serialize_bytes(&self.secret_key)
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/signing.rs
```diff
@@ -644,11 +640,28 @@ impl<'d> Deserialize<'d> for SigningKey {
...
+            fn expecting(&self, formatter: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
+                write!(
+                    formatter,
+                    concat!("An ed25519 signing (private) key")
+                )
+            }
+
+            fn visit_borrowed_bytes<E: serde::de::Error>(self, bytes: &'de [u8]) -> Result<Self::Value, E> {
+                SigningKey::try_from(bytes.as_ref()).map_err(E::custom)
+            }
```
tarcieri, line 657: Won't this fail for formats that can't borrow from their inputs, such as JSON and TOML?

It might be better to use `visit_seq`.

Also since this is a private key, it might make sense to use `serdect` which uses constant-time hex serializers for human readable formats: https://github.com/RustCrypto/formats/tree/master/serdect

ryan-mob: > Won't this fail for formats that can't borrow from their inputs, such as JSON and TOML?

oh interesting, TIL they can't borrow inputs / `visit_seq` makes sense there!

> since this is a private key, it might make sense to use serdect which uses constant-time hex serializers for human readable formats

human readable is always nice, but, that seems like it would be a change in behaviour / incompatible with binary serialisation?

tarcieri: > human readable is always nice, but, that seems like it would be a change in behaviour / incompatible with binary serialisation?

It uses binary serialization with non-human-readable formats

pinkforest:
@ryan-mob the CI is now in main - wanna rebase to that and remove `serde` from the [exclude skip list](https://github.com/dalek-cryptography/ed25519-dalek/blob/main/.github/workflows/rust.yml#L86) ?

src/signing.rs
ryan-mob: okay looking at `serdect` again, it still seems that adding this would be a breaking change to the way values are serialised compared to previous versions, which might be worthwhile (love a good human readable encoding) but, is probably outside the scope of this PR?

tarcieri: Okay, for the sake of fixing the `serde` + `no_std` serde build, we can start with this.

However, it is a breaking release, and it's really bad if people are depending on their private keys being serialized as `[12, 34, 56, ...]`.

tarcieri (approved):

src/signing.rs
ryan-mob: > However, it is a breaking release, and it's really bad if people are depending on their private keys being serialized as [12, 34, 56, ...].

yeah fully agreed on that front! and it'd be a nice usability improvement to not need to override the serde encode/decode manually for those of us who noticed / worry about these things.

ryan-mob:
thanks!

PR #293, opened 2023-03-21
Update to new `Scalar` API

rozbb:
This removes all uses of `Scalar::{from_bits, from_bits_clamped}`. Some comments inline.

`curve25519-dalek` dependency is pinned to Git in `Cargo.toml`, with revision `f460ae1` specified in `Cargo.lock`. I'll move this back to a semver dep once a new `curve25519-dalek` RC is cut.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,7 +39,11 @@ jobs:
       - run: cargo test --target ${{ matrix.target }} --features digest,rand_core
       - run: cargo test --target ${{ matrix.target }} --features serde
       - run: cargo test --target ${{ matrix.target }} --features pem
-      - run: cargo test --target ${{ matrix.target }} --all-features
+      # Can't use --all-features because of legacy_compatibility
+      - run: cargo test --target ${{ matrix.target }} --features asm,batch,digest,pem,rand_core,serde
+      - env:
+          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
+        run: cargo test --target ${{ matrix.target }} --features legacy_compatibility
 
   build-simd:
     name: Test simd backend (nightly)
@@ -82,7 +86,10 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       # No default features build
       - run: cargo build --target thumbv7em-none-eabi --release --no-default-features
-      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std
+      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std,legacy_compatibility
+      - env:
+          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
+        run: cargo build --target thumbv7em-none-eabi --release --features "legacy_compatibility"
 
   bench:
     name: Check that benchmarks compile
@@ -119,4 +126,6 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
-      - run: cargo doc --all-features
+      - env:
+          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
+        run: cargo doc --all-features
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -89,7 +89,7 @@ jobs:
       - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std,legacy_compatibility
       - env:
           RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
-        run: cargo build --target thumbv7em-none-eabi --release --features "legacy_compatibility"
+        run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features "legacy_compatibility"
 
   bench:
     name: Check that benchmarks compile
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,11 +39,7 @@ jobs:
       - run: cargo test --target ${{ matrix.target }} --features digest,rand_core
       - run: cargo test --target ${{ matrix.target }} --features serde
       - run: cargo test --target ${{ matrix.target }} --features pem
-      # Can't use --all-features because of legacy_compatibility
-      - run: cargo test --target ${{ matrix.target }} --features asm,batch,digest,pem,rand_core,serde
-      - env:
-          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
-        run: cargo test --target ${{ matrix.target }} --features legacy_compatibility
+        run: cargo test --target ${{ matrix.target }} --all-features
 
   build-simd:
     name: Test simd backend (nightly)
@@ -86,10 +82,7 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       # No default features build
       - run: cargo build --target thumbv7em-none-eabi --release --no-default-features
-      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std,legacy_compatibility
-      - env:
-          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
-        run: cargo build --target thumbv7em-none-eabi --release --no-default-features --features "legacy_compatibility"
+      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std
 
   bench:
     name: Check that benchmarks compile
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -39,7 +39,7 @@ jobs:
       - run: cargo test --target ${{ matrix.target }} --features digest,rand_core
       - run: cargo test --target ${{ matrix.target }} --features serde
       - run: cargo test --target ${{ matrix.target }} --features pem
-        run: cargo test --target ${{ matrix.target }} --all-features
+      - run: cargo test --target ${{ matrix.target }} --all-features
 
   build-simd:
     name: Test simd backend (nightly)
@@ -119,6 +119,4 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: stable
-      - env:
-          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
-        run: cargo doc --all-features
+      - run: cargo doc --all-features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ rustdoc-args = [
 features = ["batch", "pkcs8"]
 
 [dependencies]
-curve25519-dalek = { version = "=4.0.0-rc.1", default-features = false, features = ["digest"] }
+curve25519-dalek = { path = "../curve25519-dalek", default-features = false, features = ["digest"] }
 ed25519 = { version = ">=2.2, <2.3", default-features = false }
 signature = { version = ">=2.0, <2.1", optional = true, default-features = false }
 sha2 = { version = "0.10", default-features = false }
@@ -38,7 +38,7 @@ serde_bytes = { version = "0.11", optional = true }
 zeroize = { version = "1.5", default-features = false, optional = true }
 
 [dev-dependencies]
-curve25519-dalek = { version = "=4.0.0-rc.1", default-features = false, features = ["digest", "rand_core"] }
+curve25519-dalek = { path = "../curve25519-dalek", default-features = false, features = ["digest", "rand_core"] }
 hex = "0.4"
 bincode = "1.0"
 serde_json = "1.0"
@@ -64,7 +64,7 @@ batch = ["alloc", "merlin", "rand_core"]
 fast = ["curve25519-dalek/precomputed-tables"]
 digest = ["signature/digest"]
 # This features turns off stricter checking for scalar malleability in signatures
-legacy_compatibility = []
+legacy_compatibility = ["curve25519-dalek/legacy_compatibility"]
 pkcs8 = ["ed25519/pkcs8"]
 pem = ["alloc", "ed25519/pem", "pkcs8"]
 rand_core = ["dep:rand_core"]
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -64,7 +64,7 @@ batch = ["alloc", "merlin", "rand_core"]
 fast = ["curve25519-dalek/precomputed-tables"]
 digest = ["signature/digest"]
 # This features turns off stricter checking for scalar malleability in signatures
-legacy_compatibility = []
+legacy_compatibility = ["curve25519-dalek/legacy_compatibility"]
```
rozbb, line 67: `Scalar::from_bits` is necessary when we have bad scalar checks. This is why I had to make the `curve25519-dalek/legacy_compatibility` feature

src/signing.rs
```diff
@@ -688,13 +688,18 @@ impl<'d> Deserialize<'d> for SigningKey {
 // better-designed, Schnorr-based signature scheme, see Trevor Perrin's work on
 // "generalised EdDSA" and "VXEdDSA".
 pub(crate) struct ExpandedSecretKey {
+    // `key_bytes` and `key` are separate, because the public key is computed as an unreduced
+    // scalar multiplication (ie `mul_base_clamped`), whereas the signing operations are done
+    // modulo l.
+    pub(crate) key_bytes: [u8; 32],
```
rozbb, line 694: I had to add a new field to the expanded secret keys. Ideally, we'd just clamp an integer like the spec demands, and reduce modulo `l`, and let that be the canonical key. In fact, that's what [crrl](https://github.com/pornin/crrl/blob/0dd2fa2369cc8746135807ee82592f9fe07229cc/src/ed25519.rs#L1680) and [Go's ed25519](https://cs.opensource.google/go/go/+/master:src/crypto/internal/edwards25519/scalar.go;drc=da4687923b8c2d42c23f61fa3db9f4d3ce0c5f54;l=204) do.

The reason we don't do that is mainly because it is a breaking change: when `s = clamp(H(seed)[:32])` is an unreduced clamped scalar and `B` is the basepoint, then `s.reduce() * B != s * B`. We don't want people's pubkeys to change overnight, so we keep the old `s * B` computation.

src/signing.rs
```diff
@@ -707,9 +712,15 @@ impl From<&SecretKey> for ExpandedSecretKey {
         // TODO: Use bytes.split_array_ref once it’s in MSRV.
         let (lower, upper) = hash.split_at(32);
 
+        // Clamp the lower bytes. This is the key integer.
         // The try_into here converts to fixed-size array
+        let key_bytes = clamp_integer(lower.try_into().unwrap());
+        // For signing, we'll need the integer as a Scalar
+        let key = Scalar::from_bytes_mod_order(key_bytes);
```
rozbb, line 719: Previously this was unreduced. We need to ensure that this is used in a way that is consistent with how the unreduced scalar worked. Indeed this is the case. The only computation that `self.key` is used in is in `ExpandedSecretKey::{sign, sign_prehashed}`, where it's multiplied by some scalar. `Scalar::mul` is defined to work the same whether or not the inputs are reduced, so the computation is unchanged. I added the `test_mul_reduction_invariance` unit test to `curve25519-dalek` to make sure of this.

src/signature.rs
```diff
@@ -76,24 +76,14 @@ fn check_scalar(bytes: [u8; 32]) -> Result<Scalar, SignatureError> {
...
-    // significant 4 bits are unset.  If they are unset, we can succeed fast
-    // because we are guaranteed that the scalar is fully reduced.  However, if
-    // the 4th most significant bit is set, we must do the full reduction check,
-    // as the order of the basepoint is roughly a 2^(252.5) bit number.
-    //
-    // This succeed-fast trick should succeed for roughly half of all scalars.
-    if bytes[31] & 240 == 0 {
-        return Ok(Scalar::from_bits(bytes));
-    }
-
```
rozbb, line 96: I deleted this because 1) `Scalar::from_bits` doesn't exist anymore, and 2) This trick had no noticeable affect on any of the benchmarks. If this made things faster, I could have theoretically exposed `from_bits` in `curve25519-dalek` and put it behind a feature gate. Thankfully that's not necessary though.

tests/x25519.rs
```diff
@@ -16,16 +16,16 @@ fn ed25519_to_x25519_dh() {
...
-    let scalar_a = ed25519_signing_key_a.to_scalar();
-    let scalar_b = ed25519_signing_key_b.to_scalar();
+    let scalar_a = ed25519_signing_key_a.to_scalar_bytes();
+    let scalar_b = ed25519_signing_key_b.to_scalar_bytes();
 
     assert_eq!(
-        scalar_a.to_bytes(),
-        hex!("307c83864f2833cb427a2ef1c00a013cfdff2768d980c0a3a520f006904de94f")
+        scalar_a,
+        hex!("357c83864f2833cb427a2ef1c00a013cfdff2768d980c0a3a520f006904de90f")
```
rozbb, line 24: This is the unclamped hex of what's above

rozbb:
Ready for final review.

Made a small change. `ExpandedSecretKey::scalar_bytes` should be unclamped bytes. This is both in keeping with what x25519-dalek does for its secret keys, and also is what [age does](https://github.com/FiloSottile/age/blob/8e3f74c283b2e9b3cd0ec661fa4008504e536d20/agessh/agessh.go#L255-L292).

Tests are failing because we deprecated `from_bits`, and `-D warnings` doesn't like that. Gonna try to fix that. Then this is good to go.

Cargo.toml
```diff
@@ -63,9 +63,13 @@ batch = ["alloc", "merlin", "rand_core"]
...
 pkcs8 = ["ed25519/pkcs8"]
 pem = ["alloc", "ed25519/pem", "pkcs8"]
 rand_core = ["dep:rand_core"]
 serde = ["dep:serde", "ed25519/serde"]
 zeroize = ["dep:zeroize", "curve25519-dalek/zeroize"]
+
+[patch.crates-io.curve25519-dalek]
+git = "https://github.com/dalek-cryptography/curve25519-dalek.git"
+rev = "f460ae149b0000695205cc78f560d74a2d3918eb"
```
rozbb, line 75: Like x-, we'll update this when rc.3 drops.

.github/workflows/rust.yml
```diff
@@ -82,7 +86,10 @@ jobs:
       - uses: taiki-e/install-action@cargo-hack
       # No default features build
       - run: cargo build --target thumbv7em-none-eabi --release --no-default-features
-      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std
+      - run: cargo hack build --target thumbv7em-none-eabi --release --each-feature --exclude-features default,std,legacy_compatibility
+      - env:
+          RUSTFLAGS: '-D warnings -A deprecated' # legacy_compatibility triggers a deprecation warning
```
tarcieri, line 91: You can use `#[allow(deprecated)]` attributes instead of passing `-A deprecated` here, which will allow it to build without warnings

rozbb: Done. See latest 2 commits.

src/signature.rs
```diff
@@ -76,24 +80,15 @@ fn check_scalar(bytes: [u8; 32]) -> Result<Scalar, SignatureError> {
         return Err(InternalError::ScalarFormat.into());
     }
 
+    // You cannot do arithmetic with scalars construct with Scalar::from_bits. We only use this
+    // scalar for EdwardsPoint::vartime_double_scalar_mul_basepoint, which is an accepted usecase.
+    #[allow(deprecated)]
     Ok(Scalar::from_bits(bytes))
 }
 
 #[cfg(not(feature = "legacy_compatibility"))]
```
tarcieri, line 89:
```suggestion
#[cfg(not(feature = "legacy_compatibility"))]
#[allow(deprecated)]
```

rozbb: I'm confused, there's no deprecated code in the non-`legacy_compatibility` path. What does this macro do?

tarcieri: Sorry, this is the wrong place.

I was trying to suggest where to add it to fix the test failure.

rozbb: It's failing because of the intentional deprecation warning on the `legacy_compat` version of `check_scalar()`. We can either remove that, letting ed- users not see the warning, or let them see the warning  and change some flags in the CI

tarcieri: `allow(deprecated)` needs to be added here: https://github.com/dalek-cryptography/ed25519-dalek/blob/1cf99d7d06a11572f532871de583dc15c2cf79f4/src/signature.rs#L156

...or to `from_bytes`. That will fix the tests.

rozbb: Isn't that equivalent to just removing `#[deprecated(...)]` from `check_scalar()`? It's not a public function, and this is the only place it's used.

tarcieri: Oh, well in that case there's not a lot of reason to have a `#[deprecated]` attribute. Maybe just make it clear in the comments?

rozbb: Makes sense. I'll do the latter tn

rozbb:
@tarcieri ready again. Ended up not inlining `check_scalar` because the inline comments were extensive and useful. I just added doc comments to both versions.

Cargo.toml
tarcieri, line 75: FYI, this will be stored in Cargo.lock, so in the future you can do `cargo update -p curve25519-dalek` to update it without having to annotate it manually

src/signature.rs
```diff
@@ -76,24 +79,17 @@ fn check_scalar(bytes: [u8; 32]) -> Result<Scalar, SignatureError> {
...
 
+    // You cannot do arithmetic with scalars construct with Scalar::from_bits. We only use this
+    // scalar for EdwardsPoint::vartime_double_scalar_mul_basepoint, which is an accepted usecase.
+    // The `from_bits` method is deprecated because it's unsafe. We know this.
+    #[allow(deprecated)]
     Ok(Scalar::from_bits(bytes))
 }
 
+/// Ensures that the scalar `s` of a signature is within the bounds [0, ℓ)
 #[cfg(not(feature = "legacy_compatibility"))]
```
tarcieri, line 90: This might be a little easier to read if it were written with an `if` statement instead of having two definitions depending on the feature. Something like:

```rust
fn check_scalar(bytes: [u8; 32]) -> Result<Scalar, SignatureError> {
    if cfg!(feature = "legacy_compatibility") {
        [...]
    } else {
        [...]
     }
}
```

rozbb: I played with this a bit, and, because of the extensive comments, I thought it looked better the way it is currently. Is it ok if we keep it?

Cargo.toml
rozbb: Does that mean that these lines can be removed entirely?

src/signature.rs
tarcieri: Sure

Cargo.toml
tarcieri: The `rev` line, yes. You can bump it using `cargo update -p curve25519-dalek`

rozbb:
Ok should be ready to go. This also happens to fix the rare behavior change mentioned in #299 where `ExpandedSecretKey` deserialization might be different because it skipped clamping.

After this, I think this and `x-` need new passthrough feature flags `simd, simd_avx2, simd_avx512`, and that should be it for the next RC.

tarcieri:
> After this, I think this and x- need new passthrough feature flags simd, simd_avx2, simd_avx512

I'm still definitely not sold on having a bunch of features for these backends.

I feel like we managed to remove a bunch of the complexity and knobs that introduced unnecessary choices which can be made by intermediate dependencies, only to reintroduce them.

We should probably reopen one of the issues we had discussing this or make a new one to (re-)discuss backend selection. There were many, many people opposed to the use of crate features for this and I think it provides unnecessary complexity with no immediate use cases.

rozbb:
Ok agreed. When you get a chance can you approve this?

src/hazmat.rs
```diff
@@ -64,18 +68,26 @@ impl ExpandedSecretKey {
...
+    #[allow(clippy::unwrap_used)]
     pub fn from_bytes(bytes: &[u8; 64]) -> Self {
         // TODO: Use bytes.split_array_ref once it’s in MSRV.
-        let mut lower: [u8; 32] = [0u8; 32];
-        let mut upper: [u8; 32] = [0u8; 32];
+        let (lower, upper) = bytes.split_at(32);
+
+        // Convert the slices to [u8; 32]
+        let scalar_bytes = lower.try_into().unwrap();
+        let hash_prefix = upper.try_into().unwrap();
```
tarcieri, line 81: I think it'd be nice to eventually turn on the `clippy::unwrap_used` lint.

FWIW, a [quick check in godbolt](https://godbolt.org/z/sTYd48nGK) shows bounds checks are elided here.

tarcieri (approved):
This is fine but I'm not super wild about the use of `unwrap`.

rozbb:
Oh I'll change that

rozbb:
Ok I believe all the unwraps are entirely gone now

src/hazmat.rs
```diff
@@ -35,6 +35,10 @@ use curve25519_dalek::digest::{generic_array::typenum::U64, Digest};
 ///
 /// Instances of this secret are automatically overwritten with zeroes when they fall out of scope.
 pub struct ExpandedSecretKey {
+    // `scalar_bytes` and `scalar` are separate, because the public key is computed as an unreduced
+    // scalar multiplication (ie `mul_base_clamped`), whereas the signing operations are done
+    // modulo l.
+    pub(crate) scalar_bytes: [u8; 32],
```
tarcieri, line 41: I guess I overlooked this before.

So, there are use cases like Ed25519-BIP32 where the private scalar is computed rather than clamped from bytes.

Adding this field means you can no longer construct an `ExpandedSecretKey` directly from a derived `Scalar` and prefix using the struct literal syntax `ExpandedSecretKey { scalar, hash_prefix }` and have to go through one of the constructor methods.

I guess the best you could do now if you have a `scalar` and a `hash_prefix` is to serialize the scalar and concatenate it with the `hash_prefix` and use `ExpandedSecretKey::from_bytes`?

Also all that aside, I find a mixture of public and private fields weird. If you really want to go down this path it would probably be good to make `scalar` and `hash_prefix` into methods, removing the fields from `pub` visibility.

tarcieri: Opened a PR to make the other fields private: https://github.com/dalek-cryptography/curve25519-dalek/pull/544

PR #294, opened 2023-03-22
Improve diagnostics when key being deserialized is too long

robjtede:
Previous error would simply say "trailing characters" which is correct, but we can do better by counting the remaining sequence elements and displaying a nicer error "invalid length N" where N > 32, same as N < 32.


rozbb (approved):

tarcieri (approved):

rozbb:
Thank you!

PR #296, opened 2023-03-29
Add `Scalar` and `MontgomeryPoint` conversions

tarcieri:
- Adds `SigningKey::to_scalar` to extract the private scalar
- Adds `VerifyingKey::to_montgomery` to map the verifying key's `EdwardsPoint` to `MontgomeryPoint`
- ~~Also adds corresponding `From<&T>` impls which call the inherent methods.~~

This is useful for systems which are keyed using Ed25519 keys which would like to use X25519 for D-H. Having inherent methods means it's possible to call these methods without having to import `Scalar` and `MontgomeryPoint` from `curve25519-dalek`.

This is of course a bit circuitous: we could just multiply `Scalar` by `EdwardsPoint` and use the resulting `EdwardsPoint` as the D-H shared secret, however it seems many protocols have adopted this approach of mapping to `MontgomeryPoint` and using that for the shared secret, since X25519 is traditionally used for ECDH with Curve25519.

Edit: impetus for this is https://github.com/RustCrypto/nacl-compat/pull/105

That PR added the scalar extraction and Montgomery conversion via `ed25519-dalek`, but with this PR there's no need to couple through `ed25519-dalek`, instead it allows use of only the types from `curve25519-dalek`, so this provides a general solution for any crates which work in terms of `Scalar` and `MontgomeryPoint` to be keyed with Ed25519.

### DIFF
--- a/src/verifying.rs
+++ b/src/verifying.rs
@@ -424,6 +424,13 @@ impl VerifyingKey {
     ///
     /// This is useful for systems which perform X25519 Diffie-Hellman using
     /// Ed25519 keys.
+    ///
+    /// When possible, it's recommended to use separate keys for signing and
+    /// Diffie-Hellman.
+    ///
+    /// For more information on the security of systems which use the same keys
+    /// for both signing and Diffie-Hellman, see the paper
+    /// [On using the same key pair for Ed25519 and an X25519 based KEM](https://eprint.iacr.org/2021/509.pdf).
     pub fn to_montgomery(&self) -> MontgomeryPoint {
         self.point.to_montgomery()
     }
--- a/src/verifying.rs
+++ b/src/verifying.rs
@@ -18,6 +18,7 @@ use curve25519_dalek::digest::generic_array::typenum::U64;
 use curve25519_dalek::digest::Digest;
 use curve25519_dalek::edwards::CompressedEdwardsY;
 use curve25519_dalek::edwards::EdwardsPoint;
+use curve25519_dalek::montgomery::MontgomeryPoint;
 use curve25519_dalek::scalar::Scalar;
 
 use ed25519::signature::Verifier;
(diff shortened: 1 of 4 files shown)


rozbb:
A test would be good.

Also, I had to look into why people are doing this and whether it's secure. Why people are doing this: it's convenient. Age uses ed25519 ssh keys as encryption pubkeys. Fillipo wrote about it [here](https://words.filippo.io/using-ed25519-keys-for-encryption/). Is it secure: yeah, seems so. Take a look at the note at the bottom of the [libsodium page](https://libsodium.gitbook.io/doc/advanced/ed25519-curve25519).

So I'm fine with the change. I think we should copy the note from that doc page though and put it on the `to_montgomery()` method.

src/verifying.rs
```diff
@@ -471,6 +480,12 @@ where
     }
 }
 
+impl From<&VerifyingKey> for MontgomeryPoint {
```
rozbb, line 483: I'm a little concerned that this is more confusing than it is helpful. It is already very niche for someone to want to use a sig pubkey as a KEM pubkey. I think the `to_montgomery()` method is enough.

tarcieri: I can remove these. I do agree the inherent methods are more explicit.

tarcieri: Removed in 239eecb

src/verifying.rs
```diff
@@ -418,6 +419,14 @@ impl VerifyingKey {
             Err(InternalError::Verify.into())
         }
     }
+
+    /// Convert this verifying key into Montgomery form.
+    ///
+    /// This is useful for systems which perform X25519 Diffie-Hellman using
+    /// Ed25519 keys.
+    pub fn to_montgomery(&self) -> MontgomeryPoint {
```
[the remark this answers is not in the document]
tarcieri: I'm curious if this is worth documenting in the context of `VerifyingKey::to_montgomery`. It seems like if you are treating the identity point like a verifying key, something else has already gone horribly, horribly wrong.

PR #297, opened 2023-03-30
Fixed repoerted speedup/slowdown percentages in README benchmarks

rozbb:
The `fiat` column percentages were previously calculated wrt the `simd` column, and not the `u64`column, which appears to be the base.


tarcieri (approved):
utACK

Issue #164, opened 2021-03-06
Other hashing algorithms


charleschege:
Is it a good idea to add optional support for other hashing algorithms like blake3 or blake3? Or is it out of scope for this project?

tarcieri:
Ed25519 is EdDSA instantiated with SHA-512 as the hash function.

EdDSA constructions which are instantiated with other hash functions are incompatible EdDSA variants which are not "Ed25519".

charleschege:
I saw this project, https://github.com/rmw-lib/ed25519-dalek-blake2b

Does this mean that Ed25519 is by design tied to sha512 or is it just to comply with the design spec?

tarcieri:
From [RFC 8032 Section 5.1](https://tools.ietf.org/html/rfc8032#section-5.1), "Ed25519" is always instantiated with SHA-512:

```
5.1.  Ed25519ph, Ed25519ctx, and Ed25519

   Ed25519 is EdDSA instantiated with:

   +-----------+-------------------------------------------------------+
   | Parameter | Value                                                 |
   +-----------+-------------------------------------------------------+
   |     p     | p of edwards25519 in [RFC7748] (i.e., 2^255 - 19)     |
   |     b     | 256                                                   |
   |  encoding | 255-bit little-endian encoding of {0, 1, ..., p-1}    |
   |  of GF(p) |                                                       |
   |    H(x)   | SHA-512(dom2(phflag,context)||x) [RFC6234]            |
   |     c     | base 2 logarithm of cofactor of edwards25519 in       |
   |           | [RFC7748] (i.e., 3)                                   |
   |     n     | 254                                                   |
   |     d     | d of edwards25519 in [RFC7748] (i.e., -121665/121666  |
   |           | = 370957059346694393431380835087545651895421138798432 |
   |           | 19016388785533085940283555)                           |
   |     a     | -1                                                    |
   |     B     | (X(P),Y(P)) of edwards25519 in [RFC7748] (i.e., (1511 |
   |           | 22213495354007725011514095885315114540126930418572060 |
   |           | 46113283949847762202, 4631683569492647816942839400347 |
   |           | 5163141307993866256225615783033603165251855960))      |
   |     L     | order of edwards25519 in [RFC7748] (i.e.,             |
   |           | 2^252+27742317777372353535851937790883648493).        |
   |   PH(x)   | x (i.e., the identity function)                       |
   +-----------+-------------------------------------------------------+
```

[...]

PR #299, opened 2023-04-10
Add `hazmat` module with `ExpandedSecretKey`, `raw_sign`, `raw_sign_prehashed`

### LINKED ISSUE #298: Reintroduce ExpandedSecretKey funcionality
Hi there.

We're using the `dalek` crates in [vodozemac](https://github.com/matrix-org/vodozemac/) as the core for our double ratchet implementation. This is a relatively new library intended to replace our old C based [libolm](https://gitlab.matrix.org/matrix-org/olm/) library.

Now our problem is that libolm always expands Ed25519 secret keys, which are stored long-term.

As we have many users already using libolm, we would like to transfer this expanded Ed25519 key when we transition our users from libolm to vodozemac. We've looked into updating our `ed25519-dalek` dependency, but found that it is no longer possible to construct an `ExpandedSecretKey` as the type is no longer public, as per https://github.com/dalek-cryptography/ed25519-dalek/pull/205.

Is there any way to restore this functionality?

rozbb:
This splits out the signing functionality we have into two new functions. Nothing changes underlyingly. The only difference is now users can use `raw_sign()` and `raw_sign_prehashed()` if you enable the `raw_sign` feature (and `digest` feature as well for the latter).

Also this renames `ExpandedSecretKey::nonce` to `hash_prefix`, because that's what it is, and it is absolutely used more than once.

Fixes #298.

# Why expose these functions

Our removal of the public `ExpandedSecretKey` API in  #205 made it impossible for users to sign using an `ExpandedSecretKey`. The only way to sign now is `SigningKey::sign`. The issue here is that some downstream users, as noted in #298, never actually saved their `SigningKey`, and only saved the `ExpandedSecretKey`. You cannot recover a `SigningKey` from an `ExpandedSecretKey`, since the former is the seed whose hash is expanded to the latter. Thus, it is necessary to expose some lower-level way of computing signatures. There were a few levels of abstraction we could have picked here, and I'm open to suggestions.

Separately, and I don't know how important this is, but these functions would allow someone to build [BIP32-Ed25519](https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf) on top of this library, since `scalar` and `hash_prefix` are explicit inputs to the `raw_sign()` function.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ rustdoc-args = [
     "--html-in-header", "docs/assets/rustdoc-include-katex-header.html",
     "--cfg", "docsrs",
 ]
-features = ["batch", "pkcs8"]
+features = ["batch", "digest", "hazmat", "pem", "serde"]
 
 [dependencies]
 curve25519-dalek = { version = "=4.0.0-rc.2", default-features = false, features = ["digest"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,12 +62,12 @@ asm = ["sha2/asm"]
 batch = ["alloc", "merlin", "rand_core"]
 fast = ["curve25519-dalek/precomputed-tables"]
 digest = ["signature/digest"]
+# Exposes the hazmat module
+hazmat = []
 # Turns off stricter checking for scalar malleability in signatures
 legacy_compatibility = []
 pkcs8 = ["ed25519/pkcs8"]
 pem = ["alloc", "ed25519/pem", "pkcs8"]
 rand_core = ["dep:rand_core"]
-# Exposes the unsafe raw_sign() and raw_sign_prehashed() functions
-raw_sign = []
 serde = ["dep:serde", "ed25519/serde"]
 zeroize = ["dep:zeroize", "curve25519-dalek/zeroize"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ zeroize = { version = "1.5", default-features = false, optional = true }
 
 [dev-dependencies]
 curve25519-dalek = { version = "=4.0.0-rc.2", default-features = false, features = ["digest", "rand_core"] }
+blake2 = "0.10"
+sha3 = "0.10"
 hex = "0.4"
 bincode = "1.0"
 serde_json = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,10 +62,12 @@ asm = ["sha2/asm"]
 batch = ["alloc", "merlin", "rand_core"]
 fast = ["curve25519-dalek/precomputed-tables"]
 digest = ["signature/digest"]
-# This features turns off stricter checking for scalar malleability in signatures
+# Turns off stricter checking for scalar malleability in signatures
 legacy_compatibility = []
 pkcs8 = ["ed25519/pkcs8"]
 pem = ["alloc", "ed25519/pem", "pkcs8"]
 rand_core = ["dep:rand_core"]
+# Exposes the unsafe raw_sign() and raw_sign_prehashed() functions
+raw_sign = []
 serde = ["dep:serde", "ed25519/serde"]
 zeroize = ["dep:zeroize", "curve25519-dalek/zeroize"]
--- a/src/hazmat.rs
+++ b/src/hazmat.rs
@@ -66,6 +66,7 @@ impl ExpandedSecretKey {
 
     /// Construct an `ExpandedSecretKey` from an array of 64 bytes.
     pub fn from_bytes(bytes: &[u8; 64]) -> Self {
+        // TODO: Use bytes.split_array_ref once it’s in MSRV.
         let mut lower: [u8; 32] = [0u8; 32];
         let mut upper: [u8; 32] = [0u8; 32];
 
--- a/src/hazmat.rs
+++ b/src/hazmat.rs
@@ -13,7 +13,7 @@
 // defined.
 #![allow(dead_code)]
 
-use crate::{InternalError, SignatureError};
+use crate::SignatureError;
 
 use curve25519_dalek::Scalar;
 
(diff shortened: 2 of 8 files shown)


tarcieri:
My preference would be to stick the `pub` functions in a `hazmat` module with a big fat warning, similar to this:

https://docs.rs/ecdsa/latest/ecdsa/hazmat/index.html

rozbb:
That looks good. Will do

src/signing.rs
```diff
@@ -817,55 +799,196 @@ impl ExpandedSecretKey {
...
+            prehashed_message,
+            verifying_key,
+            context,
+        )
+    }
+}
 
-        let ctx: &[u8] = context.unwrap_or(b""); // By default, the context is an empty string.
+//
+// Signing funcitons. These are private to this module unless explicitly exported.
```
poljar, line 813:
```suggestion
// Signing functions. These are private to this module unless explicitly exported.
```

src/signing.rs
```diff
@@ -817,55 +799,196 @@ impl ExpandedSecretKey {
...
+///   deterministically generate the `R` part of the signature.
+///
+/// `scalar` and `hash_prefix` are usually selected such that `scalar || hash_prefix = H(sk)` where
+/// `sk` is the signing key
+#[cfg(feature = "raw-sign")]
+fn raw_sign(
+    scalar: Scalar,
+    hash_prefix: [u8; 32],
+    message: &[u8],
+    verifying_key: &VerifyingKey,
```
poljar, line 940: One thing that makes this API a bit unpleasant, but certainly not unworkable, is that we don't have a simple method to construct a `VerifyingKey` from a `Scalar`.

`VerifyingKey::clamp_and_mul_base()`, which is used by `ExpandedSecretKey`, is private as well, so we have no other choice than going through `VerifyingKey::from_bytes()`.

tarcieri: Would `impl From<EdwardsPoint> for VerifyingKey` be sufficient?

Cargo.toml
```diff
@@ -62,10 +62,12 @@ asm = ["sha2/asm"]
...
 fast = ["curve25519-dalek/precomputed-tables"]
 digest = ["signature/digest"]
-# This features turns off stricter checking for scalar malleability in signatures
+# Turns off stricter checking for scalar malleability in signatures
 legacy_compatibility = []
 pkcs8 = ["ed25519/pkcs8"]
 pem = ["alloc", "ed25519/pem", "pkcs8"]
 rand_core = ["dep:rand_core"]
+# Exposes the unsafe raw_sign() and raw_sign_prehashed() functions
+raw_sign = []
```
poljar, line 71: The `cfg!()` calls use `raw-sign`, and I think the dash is a bit nicer than the underscore for feature flags.

```suggestion
raw-sign = []
```

tarcieri: I'd suggest calling this feature `hazmat` to match a `hazmat` module.

Also in addition to having these kind of low-level signing functions, it'd be nice to have low-level verifying functions as well, which also permit customizing the digest for people who want to do that.

tarcieri:
@rozbb if you do extract a `hazmat` module it'd be good to link to https://github.com/MystenLabs/ed25519-unsafe-libs and document how things could go wrong if the public key isn't computed from the secret key

benma:
> but these functions would allow someone to build [BIP32-Ed25519](https://input-output-hk.github.io/adrestia/static/Ed25519_BIP.pdf) on top of this library

Fyi, I made such a library (you can find it [here](https://github.com/digitalbitbox/rust-bip32-ed25519)), and I would appreciate if this PR made it into 2.0, as I previously relied on `ExpandedSecretKey::from_bytes` to use the BIP32 expanded private key to sign a message.

jplatte:
@rozbb gentle ping, are you expecting to finish this soon?

Alternatively, could we maybe get a patch release on x25519-dalek with the zeroize dependency unpinned? It seems like it causes a new problem once every week 😢

rozbb:
@jplatte I apologize, I've been swamped at work lately. I will definitely get to this this week. The existing PR is pretty close to the finish line so it shouldn't take long.

Cargo.toml
rozbb: What extra knobs would raw verification have? I looked at BIP32 and its verification algorithm is unchanged from the original spec. I believe FROST's is also equivalent to the original spec (after the correct pubkey has been computed).

rozbb:
Ok, I think all concerns should be addressed, pending 1 unresolved conversation.

To address #298, I brought back `ExpandedSecretKey`. One notable change is that it uses `Scalar::from_bytes_modulo_order()` for deserialization rather than the `Scalar::from_bits()` function. Strictly speaking, this is a meaningful change. But the only way you'd notice the difference is if you were somehow serializing unreduced expanded secret key scalars. I really doubt that'll be an issue.

poljar (approved):
From my point of view this is all we need and looks great. Thanks a lot for bringing back `ExpandedSecretKey` as well.

Cargo.toml
rozbb: Oh you mean make  the non-prehashed variant to generic over a 512-bit hash function? Is this something people asked for?

tarcieri: Yes, e.g. #164

rozbb: Ah ok. Then maybe `raw_sign` should be generic too?

rozbb:
Alright, @tarcieri ready to review. Sorry for the huge edits. This weirdly affected a lot of things.

src/hazmat.rs
```diff
@@ -0,0 +1,260 @@
...
+        bytes
+    }
+
+    /// Construct an `ExpandedSecretKey` from a slice of bytes.
+    ///
+    /// # Returns
+    ///
+    /// A `Result` whose okay value is an EdDSA `ExpandedSecretKey` or whose error value is an
+    /// `SignatureError` describing the error that occurred.
+    pub fn from_bytes(bytes: &[u8]) -> Result<ExpandedSecretKey, SignatureError> {
```
tarcieri, line 73: Elsewhere (e.g. in `curve25519-dalek`, but also in @RustCrypto crates including `ed25519::Signature`) we adopted a convention that `from_bytes` accepts a byte array and `from_slice` accepts a byte slice (with a `TryFrom<&[u8]>` impl that thunks to `from_slice`.

Notably this allows `from_bytes` to be infallible.

```suggestion
    pub fn from_bytes(bytes: &[u8; 64]) -> Self {
```

tarcieri:
One nit on `from_bytes` accepting a slice but otherwise this looks good to me.

src/hazmat.rs
```diff
@@ -0,0 +1,248 @@
...
+        let mut bytes: [u8; 64] = [0u8; 64];
+
+        bytes[..32].copy_from_slice(self.scalar.as_bytes());
+        bytes[32..].copy_from_slice(&self.hash_prefix[..]);
+        bytes
+    }
+
+    /// Construct an `ExpandedSecretKey` from an array of 64 bytes.
+    pub fn from_bytes(bytes: &[u8; 64]) -> Self {
+        let mut lower: [u8; 32] = [0u8; 32];
```
tarcieri, line 69: Maybe a TODO comment here for `split_array`(_ref) like we have elsewhere?

```suggestion
        // TODO: Use bytes.split_array_ref once it’s in MSRV.
        let mut lower: [u8; 32] = [0u8; 32];
```

tarcieri (approved):
Looks good now, although for bonus points an `ExpandedSecretKey::from_slice` method and/or `TryFrom<&[u8]>` impl would be nice.

Cargo.toml
tarcieri: Looks like it's now generic around `CtxDigest` so we should be good

rozbb:
Good idea, thanks!

jplatte:
Any plans for the next (pre-)release? :)

benma:
> Any plans for the next (pre-)release? :)

Seconded :)

tarcieri:
It's blocked on https://github.com/dalek-cryptography/curve25519-dalek/pull/531

tarcieri:
It's out

PR #301, opened 2023-04-13
Rename `ExpandedSecretKey::nonce` => `prefix`

tarcieri:
The name "nonce" doesn't really capture the role of this value, which unfortunately seems to resist a pithy name across all of the specs including RFC8032, which refers to it as `h_b, ..., h_(2b-1)`. A nonce is used once whereas this value is used repeatedly for every message signed.

Its usage is a prefix concatenated onto the message to be hashed and used to compute `r`, so given that, this commit renames it to `prefix`.

### DIFF
--- a/src/signing.rs
+++ b/src/signing.rs
@@ -737,14 +737,14 @@ impl<'d> Deserialize<'d> for SigningKey {
 // "generalised EdDSA" and "VXEdDSA".
 pub(crate) struct ExpandedSecretKey {
     pub(crate) scalar: Scalar,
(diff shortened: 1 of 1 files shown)


burdges:
It's clearly not a nonce.  I've called it nonce_seed but yeah prefix.. lol

rozbb:
Is this instead of #299? I've renamed it `hash_prefix` there. 

Sorry haven't gotten to anything this week. Been very busy

tarcieri:
Oh whoops, so you did

PR #303, opened 2023-05-16
Attempt at implementing streaming verification

Monadic-Cat:
As requested [here](https://github.com/dalek-cryptography/ed25519-dalek/pull/196#issuecomment-1550194521), this PR contains my port of PR #196 to use the `hazmat` module instead of implementing any cryptographic *stuff* itself.

I haven't copied over the added test yet, but I have confirmed basic functionality by test using it in a demo application I have lying around.

### DIFF
--- a/src/streaming_verifying.rs
+++ b/src/streaming_verifying.rs
@@ -1,0 +1,172 @@
+use crate::{SignatureError, VerifyingKey};
+use crate::Signature;
+use crate::hazmat;
+use sha2::{Sha512, Digest};
+
+pub struct StreamingVerifier {
+    /// Public key to verify with.
+    pub(crate) public_key: VerifyingKey,
+
+    /// Candidate signature to verify against.
+    pub(crate) signature: Signature,
+
+    /// Hash state.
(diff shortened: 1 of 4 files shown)


tarcieri:
Thanks for pushing this up. It's interesting you managed to make it work but I guess I misjudged whether this would be a good fit. Feels like the wrong abstraction level.

Perhaps `VerifyingKey::recompute_R` can be subdivided into some smaller functions (that is to say, preserve `recompute_R`, but have it call some other functions) which are more amenable to this streaming verification use case.

e.g. `recompute_R_from_k`?

src/streaming_verifying.rs
```diff
@@ -0,0 +1,172 @@
+use crate::{SignatureError, VerifyingKey};
```
tarcieri, line 1: Perhaps this module could be `src/verifying/streaming.rs`?

tarcieri:
Looks like @robjtede updated #196 so perhaps we should continue the discussion there

tarcieri:

PR #304, opened 2023-06-12
Sign/verify by digest update, StreamVerifier refactoring

mkj:
I'd like to be able to sign/verify non-prehash signatures without the whole message in memory. The use case is for running on `no_std` embedded where the message is serialized directly into the sha512 digest. It's for SSH protocol so I can't use ed25519 prehashed.

There's the existing work in #196 for `StreamVerifier`, but that can't be extended to signing since signing needs two passes. Instead I've added `raw_sign_byupdate()` and `raw_verify_byupdate()` that take a closure to update the message digest.

As suggested in https://github.com/dalek-cryptography/ed25519-dalek/pull/196#discussion_r1195798967 I've moved `recompute_R` into its own struct and made `StreamVerifier` use it too. 

Would something like this be reasonable to add? I'm open to other suggestions on how achieve the same thing.

(I only just noticed `ComputeR` is a bit of a silly looking name, that could change. Likewise `_byupdate` isn't the best name, but other names I came up with would get confused with prehashed methods).


tarcieri:
This crate has an interesting API for signing incrementally: https://docs.rs/ec25519/latest/ec25519/struct.SecretKey.html#method.sign_incremental

tarcieri:

PR #306, opened 2023-06-18
ConstantTimeEq and PartialEq for SigningKey

mkj:
My use case for it is as part of a round-trip test for serialising a SigningKey inside a program's config struct. I don't see any hazard adding this as a constant time comparison?

I've added `subtle` at 2.3.0. the same as curve25519-dalek.


tarcieri (approved):

coltfred:
It seems that this could be closed since it's actually already done in the other PR?

tarcieri:

